PR #37, opened 2021-04-14
[DM-28120] Support Gafaelfawr 2.0

rra:
Allow a Gafaelfawr admin token to be passed in via an environment
variable as configuration.  If that is set, get a user token for
the target user via a Gafaelfawr admin API call.  If not, fall back
on the old JWT construction logic.

### DIFF
--- a/src/mobu/user.py
+++ b/src/mobu/user.py
@@ -1,5 +1,7 @@
 """All the data for a mobu user."""
 
+from __future__ import annotations
+
 __all__ = [
     "User",
 ]
@@ -11,6 +13,7 @@ from dataclasses import dataclass
 from string import Template
 
 import jwt
+from aiohttp import ClientSession
 
 from mobu.config import Configuration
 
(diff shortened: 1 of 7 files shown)


src/mobu/user.py
```diff
@@ -21,12 +24,37 @@ class User:
...
+        admin_token = Configuration.gafaelfawr_token
+        async with ClientSession() as s:
+            r = await s.post(
+                token_url,
+                headers={"Authorization": f"bearer {admin_token}"},
+                json={
+                    "username": username,
+                    "token_type": "user",
+                    "token_name": f"mobu {str(int(time.time()))}",
+                    "scopes": ["exec:notebook"],
```
cbanek, line 47: we might need to add more scopes here, since I think this token wll also be used for things like TAP queries from the notebook, but this is fine for now

cbanek (approved):

src/mobu/user.py
rra: Want me to make this a configuration parameter? It wouldn't be too hard to do while I'm in there, and then we could plumb it through the Helm chart and add whatever scopes we need via `values.yaml`.

cbanek: I think it's fine for now, although we probably want to be able later on to configure it from the body of the POST that contains other options (like the URL or the name of the image) since we might want a bit of a mix of scopes going on.  Like for monkeys just running against TAP directly.  Mostly just wondering if this will maybe cause a downstream failure that I have to watch for, but it'll probably be fine.  I think this token won't be used anymore anyway with nublado2, since nublado2 I think will get another token and use that instead?

rra: Oh, yeah, that makes sense. I'll go with this for now.

This is the token that will be used with nublado2. mobu gets a token for mobu that has the `admin:token` scope, which allows it to mint arbitrary tokens, and this is the code where it mints a token for a user to use to run the job. So it has to say what scopes that user token is supposed to have.

PR #44, opened 2021-04-26
[DM-29799] Pass in list of scopes from request

cbanek:
This allows for each user to have a set of scopes related to
the work it is doing, rather than a hardcoded list.

### DIFF
--- a/scripts/nublado2/lsptestuser05.json
+++ b/scripts/nublado2/lsptestuser05.json
@@ -2,7 +2,17 @@
   "name": "Robo-Simon",
   "user": {
     "username": "lsptestuser05",
-    "uidnumber": 60185
+    "uidnumber": 60185,
+    "scopes": [
+        "read:workspace/user",
+        "write:workspace/user",
+        "exec:user",
+        "read:tap",
+        "read:image",
+        "exec:portal",
+        "exec:notebook",
+        "read:workspace"
+    ]
   },
   "business": "NotebookRunner",
   "options": {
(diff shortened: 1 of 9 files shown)


scripts/nublado2/lsptestuser05.json
```diff
@@ -2,7 +2,17 @@
...
+    "uidnumber": 60185,
+    "scopes": [
+        "read:workspace/user",
+        "write:workspace/user",
+        "exec:user",
+        "read:tap",
+        "read:image",
+        "exec:portal",
+        "exec:notebook",
+        "read:workspace"
```
rra, line 14: This will fail if used against a live Gafaelfawr because most of those scopes aren't valid. The valid scopes are:

https://github.com/lsst-sqre/charts/blob/master/charts/gafaelfawr/values.yaml#L140

You probably don't want `admin:token` or `user:token`.

cbanek: Ah fantastic, thanks, yeah I was copying the list from the current list in the old tokens.

cbanek: Just wondering, did we get rid of the image scope too it seems?  I think this was used by SODA, but I'm not sure what it is using now, since it's not in the normal flow of things

rra (approved):
LGTM

scripts/nublado2/lsptestuser05.json
rra: I dropped all the scopes that nothing was using. I can definitely add them back in (it's a simple change to the Gafaelfawr chart), but was leaning towards adding them in the same change that adds the access control rule so that we don't accumulate unused stuff like we were before.

cbanek: Sounds right!

PR #53, opened 2021-06-14
[neophile] Update dependencies

sqrbot:
- Update frozen Python dependencies
- Update psf/black pre-commit hook from 21.5b2 to 21.6b0


rra:
Will handle this as part of Adam's PR.

PR #63, opened 2021-07-13
Shut down kernel after notebook execution

athornton:


rra (commented):
Don't you need to do this in `NotebookRunner`? This seems to change only the simple loop.

athornton:
> Don't you need to do this in `NotebookRunner`? This seems to change only the simple loop.

No, I'm not putting a count in NotebookRunner, I'm just doing a new kernel per complete notebook.  For the not-in-a-notebook executor I can't do that, hence MAX_EXECUTIONS.

rra (commented):
I'm confused, then, because I don't think we've seen the problem with this runner, and we've definitely seen the problem with the `NotebookRunner`. So I'm not sure how this change can address the problem we're trying to fix.

rra (approved):
Oh, wait, never mind, I see, you're adding a delete to `NotebookRunner`. Somehow I missed that.

PR #65, opened 2021-07-15
Add guard/warning/retry around 403s from Jupyter kernel

athornton:


athornton:
The retry doesn't actually help, because it really isn't transient, so this adds complexity for little gain.  Closing.

PR #67, opened 2021-07-19
Retry hub login and then dispatch request again when we get an auth error

athornton:
This attempts to address the 403 errors we've been seeing.


athornton:
closing, superseded by subsequent events

PR #70, opened 2021-07-23
[DM-31083] Add autostart support to mobu

rra:
- Switch to Pydantic models for the API and internal data passing
- Overhaul the REST API to group monkeys into flocks for easier management
- Add support for autostarting flocks of monkeys via a YAML configuration file


src/mobu/models/business.py
```diff
@@ -0,0 +1,85 @@
+"""Models for monkey business."""
```
athornton, line 1: I clearly need to learn something about Pydantic.  Models look a lot like dataclasses to me.  Ah, I see, you can sort of substitute dataclasses for BaseModels: https://pydantic-docs.helpmanual.io/usage/dataclasses/

src/mobu/business/jupyterloginloop.py
```diff
@@ -31,10 +30,13 @@ class JupyterLoginLoop(Business):
...
-        self, logger: BoundLogger, options: Dict[str, Any], user: User
+        self,
+        logger: BoundLogger,
+        business_config: BusinessConfig,
+        user: AuthenticatedUser,
     ) -> None:
-        super().__init__(logger, options, user)
-        self._client = JupyterClient(user, logger, options)
+        super().__init__(logger, business_config, user)
+        self._client = JupyterClient(user, logger, business_config)
```
athornton, line 39: Something I've been curious about reading your code; usually when doing parameters, even if they can be positional, I name them.  Am I doing it wrong, or is this simply a matter or personal preference?

src/mobu/business/notebookrunner.py
```diff
@@ -17,35 +17,37 @@
...
+from ..models.business import BusinessData
 from .jupyterloginloop import JupyterLoginLoop
 
 if TYPE_CHECKING:
     from typing import Any, Dict, Iterator, List, Optional
 
     from structlog import BoundLogger
 
-    from ..user import User
+    from ..models.business import BusinessConfig
```
athornton, line 28: I like this much better than the old options.

src/mobu/dependencies/manager.py
```diff
@@ -23,7 +18,7 @@ class MonkeyBusinessManager:
     """Manages all of the running monkeys."""
 
     def __init__(self) -> None:
-        self._monkeys: Dict[str, Monkey] = {}
+        self._flocks: Dict[str, Flock] = {}
```
athornton, line 21: So the base unit of MonkeyBusiness is no longer a single monkey, but instead a flock?

src/mobu/flock.py
```diff
@@ -0,0 +1,114 @@
...
+            assert self._config.user_spec
+            count = self._config.count
+            users = self._users_from_spec(self._config.user_spec, count)
+        scopes = self._config.scopes
+        return [
+            await AuthenticatedUser.create(u, scopes, self._session)
+            for u in users
+        ]
+
+    def _users_from_spec(self, spec: UserSpec, count: int) -> List[User]:
```
athornton, line 106: Knowing about zfill() would have made my formatting code less gross, for sure.

src/mobu/flock.py
```diff
@@ -0,0 +1,114 @@
+"""A flock of monkeys doing business."""
```
athornton, line 1: I like this.

src/mobu/autostart.py
```diff
@@ -0,0 +1,26 @@
+"""Support for automatically starting flocks."""
```
athornton, line 1: Very straightforward.

src/mobu/handlers/external.py
```diff
@@ -15,9 +16,22 @@
     MonkeyBusinessManager,
     monkey_business_manager,
 )
-from ..models import Index
-
-__all__ = ["external_router", "get_index", "get_users", "post_user"]
+from ..models.error import ErrorModel
```
athornton, line 19: You might want a convenience method to get logs for all monkeys in a flock without having to manually walk the list...but maybe not.

src/mobu/jupyterclient.py
```diff
@@ -17,21 +17,18 @@
 from aiohttp import ClientResponse, ClientSession, TCPConnector
 
 from .config import config
-from .user import User
```
athornton, line 20: This will still be easy to rebase my changes onto.

athornton (approved):
You probably want Jonathan's eyes on this too, but I in general like it very much.  It will of course require retooling of monkeyflocker and fly_my_monkeys.sh but I can do both of those.  The Flock as the basic organizational unit is a very welcome change.

src/mobu/flock.py
```diff
@@ -81,9 +82,14 @@ async def start(self) -> None:
             await monkey.start(self._scheduler)
 
     async def stop(self) -> None:
-        """Stop all the monkeys."""
-        for monkey in self._monkeys.values():
-            await monkey.stop()
+        """Stop all the monkeys.
```
athornton, line 85: This isn't going to work.  At least, that was my experience trying this with gather.

At the very least you are going to want to keep track of which monkeys correctly shut down the first time, and iterate on that quickly-shrinking list.  You will be catching a number of exceptions if you fly a few dozen or more monkeys at once.

rra: If this doesn't work, there's a bug in stop (and indeed there are several problems with how stop works right now, such as the fact that it's entirely disconnected from the running event loop and races with it). I don't want to hide those bugs; I want to fix them properly.

src/mobu/dependencies/manager.py
rra: Right, which can be a single monkey.

src/mobu/business/jupyterloginloop.py
rra: It's a matter of personal taste. I tend to use the rule of three: if there are more than three parameters, they should probably be named.

src/mobu/models/business.py
rra: Pydantic models are a lot like dataclasses except they have validation support, which is necessary for FastAPI.

rra:
Okay, monkeyflocker update is now also included in this. Remaining work that I know of, apart from the charts changes, is to fix how stop works.

cbanek (approved):
Looks great to me!  Ship it!

PR #72, opened 2021-07-27
[DM-31083] Use PUT instead of POST for creating flocks

rra:
The client uploads a complete configuration, not a partial one
that's modified, so PUT seems like a more appropriate verb.


athornton (approved):
I am rarely so fussy about HTTP methods and I always use a POST, but you are probably aesthetically correct.  Looks fine.

PR #73, opened 2021-07-27
Use JupyterLab sessions rather than kernels

athornton:
Also: make timing configurable, freshen deps, rename jupyterhub->jupyter (since it also does Lab functions)


src/mobu/business/notebookrunner.py
```diff
@@ -113,10 +113,11 @@ async def lab_business(self) -> None:
 
             for cell in cells:
                 self.running_code = "".join(cell["source"])
-                await self.execute_code(kernel, self.running_code)
+                await self.execute_code(session, self.running_code)
 
         self.running_code = None
-        await self.delete_kernel(kernel)
+        await self.idle()
```
rra, line 119: Not sure why you're adding an idle here. It feels like we should be able to close the session immediately after executing the code?

src/mobu/business/notebookrunner.py
```diff
@@ -129,23 +130,30 @@ def read_notebook(self, name: str, path: str) -> List[Dict[str, Any]]:
...
-    async def create_kernel(self) -> str:
-        self.logger.info("create_kernel")
-        with self.timings.start("create_kernel"):
-            kernel = await self._client.create_kernel()
-        return kernel
+    async def create_session(self) -> JupyterLabSession:
+        self.logger.info("create_session")
+        notebook_name = ""
+        if self.notebook:
+            notebook_name = self.notebook.name
```
rra, line 137:
```suggestion
        notebook_name = self.notebook.name if self.notebook else None
```

src/mobu/jupyterclient.py
```diff
@@ -212,42 +228,75 @@ async def delete_lab(self) -> None:
             if r.status not in [200, 202, 204]:
                 await self._raise_error("Error deleting lab", r)
 
-    async def create_kernel(self, kernel_name: str = "LSST") -> str:
-        kernel_url = (
-            self.jupyter_url + f"user/{self.user.username}/api/kernels"
+    async def create_labsession(
+        self, kernel_name: str = "LSST", notebook_name: str = ""
```
rra, line 232:
```suggestion
        self, kernel_name: str = "LSST", notebook_name: Optional[str] = None
```

src/mobu/jupyterclient.py
```diff
@@ -212,42 +228,75 @@ async def delete_lab(self) -> None:
...
+    ) -> JupyterLabSession:
+        session_url = (
+            self.jupyter_url + f"user/{self.user.username}/api/sessions"
         )
-        body = {"name": kernel_name}
+        session_type = "console"
+        if notebook_name != "":
+            session_type = "notebook"
```
rra, line 239:
```suggestion
        session_type = "notebook" if notebook_name else "console"
```

src/mobu/jupyterclient.py
```diff
@@ -31,6 +37,16 @@
...
 
 
+@dataclass(frozen=True)
+class JupyterLabSession:
+    """This holds the information a client needs to talk to the Lab in order
+    to execute code."""
+
+    session_id: str = ""
+    kernel_id: str = ""
+    websocket: Optional[ClientWebSocketResponse] = None
```
rra, line 47: I don't think any of these should have default values. You always pass in all three, so they can be mandatory.

src/monkeyflocker/client.py
```diff
@@ -67,25 +67,25 @@ async def start(self, spec_file: Path) -> None:
         assert self._session, "Must be used as a context manager"
         with spec_file.open("r") as f:
             spec = yaml.safe_load(f)
-        self._logger.info("Starting flock %s", spec["name"])
+        self._logger.info(f"Starting flock {spec['name']}")
         url = urljoin(self._base_url, "/mobu/flocks")
         await self._session.put(url, json=spec)
-        self._logger.info("Flock %s started", spec["name"])
+        self._logger.info(f"Flock {spec['name']} started")
```
rra, line 73: Why these changes? I was using `%s` formatting to avoid the nested quote types.

src/mobu/business/jupyterloginloop.py
```diff
@@ -77,8 +77,7 @@ async def lab_business(self) -> None:
 
         Placeholder function intended to be overridden by subclasses.
         """
-        with self.timings.start("lab_wait"):
-            await asyncio.sleep(60)
+        await self.idle()
```
rra, line 80: I don't think these should be unified into one idle. I would keep the pause between cells as a separate configuration option than the pause between cycles of the lab. This is important for the test suite where you still want a long pause between cycles (since we run it for one cycle and then stop it), but you want to set the pause between cells to 0.

tests/business/jupyterpythonloop_test.py
```diff
@@ -28,6 +28,7 @@ async def test_run(
             "count": 1,
             "user_spec": {"username_prefix": "testuser", "uid_start": 1000},
             "scopes": ["exec:notebook"],
+            "options": {"idle_time": 2, "settle_time": 2, "max_executions": 2},
```
rra, line 31: If you make the change mentioned above, you can, here, set the settle time and pause between cells to 0 but leave the idle time at the end of each cycle at the default, and that will work well with the test and speed it up even further.

tests/conftest.py
```diff
@@ -59,15 +59,19 @@ async def client(app: FastAPI) -> AsyncIterator[AsyncClient]:
...
 
     # aioresponses has no mechanism to mock ws_connect, so we can't properly
     # test JupyterClient.run_python.  For now, just mock it out entirely.
     with patch.object(JupyterClient, "run_python") as mock:
         mock.return_value = "4"
-        yield
+        # Same problem, but up a layer now that we're using sessions and
+        # reusing the websocket.
+        with patch.object(JupyterClient, "_websocket_connect") as mock2:
+            mock2.return_value = None
```
rra, line 73: If you return `MagicMock()` instead, you can get rid of the `Optional` on this value, which in turn simplifies some other code and removes the need for the `None` check.

```suggestion
            mock2.return_value = MagicMock()
```

and import `MagicMock` from `unittest.mock`.

tests/support/jupyter.py
```diff
@@ -19,56 +20,56 @@
...
 
     This should be invoked via mocked HTTP calls so that tests can simulate
-    making REST calls to the real JupyterHub.
+    making REST calls to the real JupyterHub/Lab.
     """
 
     def __init__(self) -> None:
-        self.kernels: Dict[str, str] = {}
-        self.state: Dict[str, JupyterHubState] = {}
+        self.sessions: Dict[str, Any] = {}
```
rra, line 47:
```suggestion
        self.sessions: Dict[str, JupyterLabSession] = {}
```

src/mobu/business/notebookrunner.py
athornton: I think I was just keeping the same structure as our other loops?  We don't need it, and I'll remove it.

src/monkeyflocker/client.py
athornton: When I run the `%s` version in my development environment I get an error about wrong number of arguments.  Maybe the logger changes between Python 3.8 and 3.9 or something, and I am running one and you're running the other?

tests/conftest.py
athornton: Alas, no, then I get 
    await session.websocket.close()
TypeError: object MagicMock can't be used in 'await' expression
ERROR    testuser1:monkey.py:79 [error    ] Slack Alert: 2021-07-27 13:32:06 testuser1 object MagicMock can't be used in 'await' expression [testuser1]
﻿

athornton: But there's AsyncMock!

src/mobu/business/notebookrunner.py
athornton: done

src/mobu/jupyterclient.py
athornton: Done

src/mobu/business/jupyterloginloop.py
athornton: Done

tests/conftest.py
athornton: AsyncMock works.  Done!

tests/support/jupyter.py
athornton: Done

tests/business/jupyterpythonloop_test.py
athornton: Split lab_idle and execution_idle apart.

src/mobu/jupyterclient.py
athornton: Done

src/monkeyflocker/client.py
rra: Oh, weird. Okay, in that case, looks good to me.

rra (approved):

PR #74, opened 2021-07-29
[DM-31201] Clean shutdown and other fixes

rra:
- Add proper shutdown support for monkeys using a queue to coordinate
- Refactor the business classes more to reuse more code and better document overrides
- Make shutting down the lab between notebook executions optional
- Use max_interations for NotebookRunner as well, instead of its own config variable
- Lift more infrastructure into the base business class
- Reauthenticate to the lab before deleting a session
- Poll the hub URL instead of the pending spawn URL to see if the lab is up since the latter sometimes 503s
- Make the default JupyterPythonLoop code produce cleaner logs
- Fix desynchronization of the web socket that caused the results of code execution to be lost
- Remove the periodic hub reauth code (should now be unnecessary)
- Wait for the lab to shut down after deletion
- Be more precise when checking if the lab is running

### DIFF
--- a/src/mobu/business/base.py
+++ b/src/mobu/business/base.py
@@ -36,7 +36,8 @@ class Business:
 
     Subclasses should override ``startup``, ``execute``, and ``shutdown`` to
     add appropriate behavior.  ``idle`` by default waits for ``idle_time``,
-    which generally does not need to be overridden.
+    which generally does not need to be overridden.  Subclasses should also
+    override ``close`` to shut down any object state created in ``__init__``.
 
     All delays should be done by calling ``pause``, and the caller should
     check ``self.stopping`` and exit any loops if it is `True` after calling
@@ -67,26 +68,33 @@ class Business:
         self.control: Queue[BusinessCommand] = Queue()
         self.stopping = False
 
+    async def close(self) -> None:
+        """Clean up any business state on shutdown."""
+        pass
+
     async def run(self) -> None:
         """The core business logic, run in a background task."""
         self.logger.info("Starting up...")
-        await self.startup()
+        try:
+            await self.startup()
 
-        while not self.stopping:
-            self.logger.info("Starting next iteration")
-            try:
-                await self.execute()
-                self.success_count += 1
-            except Exception:
-                self.failure_count += 1
-                raise
-            await self.idle()
+            while not self.stopping:
+                self.logger.info("Starting next iteration")
+                try:
+                    await self.execute()
+                    self.success_count += 1
+                except Exception:
+                    self.failure_count += 1
+                    raise
+                await self.idle()
 
-        self.logger.info("Shutting down...")
-        await self.shutdown()
-
-        # Tell the control channel we've processed the stop command.
-        self.control.task_done()
+            self.logger.info("Shutting down...")
+            await self.shutdown()
+            await self.close()
+        finally:
+            # Tell the control channel we've processed the stop command.
+            if self.stopping:
+                self.control.task_done()
 
     async def startup(self) -> None:
         """Run before the start of the first iteration and then not again."""
@@ -108,9 +116,11 @@ class Business:
 
     async def stop(self) -> None:
         """Tell the running background task to stop and wait for that."""
+        self.stopping = True
         await self.control.put(BusinessCommand.STOP)
         await self.control.join()
         self.logger.info("Stopped")
+        await self.close()
 
     async def pause(self, seconds: float) -> None:
         """Pause for up to the number of seconds, handling commands."""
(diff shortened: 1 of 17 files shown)


athornton (approved):
Wow, this is a lot.  It all seems sensible enough, although I worry that I haven't really wrapped my head around it entirely.

src/mobu/business/notebookrunner.py
```diff
@@ -30,7 +29,7 @@
 __all__ = ["NotebookRunner"]
 
 
-class NotebookRunner(JupyterLoginLoop):
+class NotebookRunner(JupyterPythonLoop):
```
athornton, line 32: Glad you did this refactor.  I had it on my queue.

src/mobu/jupyterclient.py
```diff
@@ -217,17 +227,30 @@ async def spawn_lab(self) -> None:
...
             self.jupyter_url + f"hub/api/users/{self.user.username}/server"
         )
         self.log.info(f"Deleting lab for {self.user.username} at {server_url}")
-
+        headers = {"Referer": self.jupyter_url + "hub/home"}
         async with self.session.delete(server_url, headers=headers) as r:
             if r.status not in [200, 202, 204]:
                 await self._raise_error("Error deleting lab", r)
 
+        # Wait for the lab to actually go away.  If we don't do this, we may
```
athornton, line 242: Hooray!

src/mobu/models/business.py
```diff
@@ -61,39 +51,72 @@ class BusinessConfig(BaseModel):
     settle_time: int = Field(
         10,
         title="How long to wait after lab creation in seconds",
-        description="Only used by the NotebookRunner",
+        description=(
+            "Wait this long after lag creation before trying to use the lab"
```
athornton, line 55: "lab"

src/mobu/models/business.py
```diff
@@ -61,39 +51,72 @@ class BusinessConfig(BaseModel):
...
     max_executions: int = Field(
         25,
-        title="How many cells to execute in a given kernel session",
-        description="Only used by JupyterPythonLoop",
+        title="How much to execute in a given lab and session",
+        description=(
+            "For JupyterPythonLoop, this is the number of code snippets to"
+            " execute before restarting the lab. For NotebookRunner, it's"
+            " the number of notebooks."
+        ),
```
athornton, line 105: I suppose I could answer this myself, but does this turn into "number of notebooks times iterations-per-notebook" ?  Or is it "iterations of a notebook" ?

tests/conftest.py
```diff
@@ -33,21 +33,30 @@ def configure() -> Iterator[None]:
...
 
 @pytest.fixture
-async def app() -> AsyncIterator[FastAPI]:
+async def app(jupyter: MockJupyter) -> AsyncIterator[FastAPI]:
     """Return a configured test application.
 
     Wraps the application in a lifespan manager so that startup and shutdown
     events are sent during test execution.
+
+    Notes
```
athornton, line 50: Ooooh, that's what was going on there.

src/mobu/business/base.py
```diff
@@ -17,9 +19,30 @@
...
+    - Run ``startup``
+    - In a loop, run ``execute`` followed by ``idle`` until told to stop
+    - When told to stop, run ``shutdown``
+
+    Subclasses should override ``startup``, ``execute``, and ``shutdown`` to
+    add appropriate behavior.  ``idle`` by default waits for ``idle_time``,
+    which generally does not need to be overridden.  Subclasses should also
+    override ``close`` to shut down any object state created in ``__init__``.
+
+    All delays should be done by calling ``pause``, and the caller should
```
athornton, line 42: Relying on the caller to do this on each delay still feels gross, but it's definitely a step forward.

src/mobu/models/business.py
rra: I have now replaced this with:
```python
        description=(
            "For JupyterPythonLoop, this is the number of code snippets to"
            " execute before restarting the lab. For NotebookRunner, it's"
            " the number of complete notebooks. NotebookRunner goes through"
            " the directory of notebooks one-by-one, running the entirety"
            " of each one and starting again at the beginning of the list"
            " when it runs out, until it has executed a total of"
            " max_executions notebooks. It then closes the session (and"
            " optionally deletes and recreates the lab, controlled by"
            " delete_lab), and then picks up where it left off."
        ),
```

src/mobu/business/base.py
rra: Yeah, I was trying to come up with a good alternative but none of them felt very satisfying. I think the next best option would be to raise an exception that's caught by the base business class, but there will still be business-specific cleanup that has to be done when shutting down, so there will need to be a business method that's then called to do that. Although I guess that's `shutdown`, so maybe that's enough. Maybe I'll experiment with that, although it does feel like using exceptions for normal control flow, which is something you're not supposed to do.

PR #75, opened 2021-07-29
[DM-31201] Use a different approach for detecting lab status

rra:
- Redo detection of whether the lab is running and whether it spawned
- Rationalize logging and remove some unnecessary lines
- Clean up typing for the web socket connection

### DIFF
--- a/src/mobu/business/jupyterloginloop.py
+++ b/src/mobu/business/jupyterloginloop.py
@@ -59,6 +59,7 @@ class JupyterLoginLoop(Business):
         await self.lab_settle()
         if self.stopping:
             return
+        await self.lab_login()
         await self.lab_business()
         if self.config.delete_lab:
             await self.hub_login()
@@ -75,11 +76,14 @@ class JupyterLoginLoop(Business):
     async def ensure_lab(self) -> None:
         with self.timings.start("ensure_lab"):
             await self._client.ensure_lab()
-        self.logger.info("Lab created")
 
     async def lab_settle(self) -> None:
         with self.timings.start("lab_settle"):
             await self.pause(self.config.settle_time)
+
+    async def lab_login(self) -> None:
+        with self.timings.start("lab_login"):
+            await self._client.lab_login()
 
     async def delete_lab(self) -> None:
         self.logger.info("Deleting lab")
--- a/src/mobu/jupyterclient.py
+++ b/src/mobu/jupyterclient.py
@@ -11,40 +11,40 @@ import random
 import re
 import string
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from http.cookies import BaseCookie
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
-from aiohttp import (
-    ClientResponse,
-    ClientSession,
-    ClientWebSocketResponse,
-    TCPConnector,
-)
+from aiohttp import ClientResponse, ClientSession, TCPConnector
 
 from .config import config
-from .exceptions import NotebookException
+from .exceptions import LabSpawnTimeoutError, NotebookException
 
 if TYPE_CHECKING:
-    from typing import Any, Optional
+    from typing import Any, NoReturn, Optional
 
+    from aiohttp import ClientWebSocketResponse
     from aiohttp.client import _RequestContextManager, _WSRequestContextManager
     from structlog import BoundLogger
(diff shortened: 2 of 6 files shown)


src/mobu/business/jupyterloginloop.py
```diff
@@ -59,6 +59,7 @@ async def execute(self) -> None:
         await self.lab_settle()
         if self.stopping:
             return
+        await self.lab_login()
```
athornton, line 62: Is it this that ensures we have the right token/cookies ?

src/mobu/jupyterclient.py
```diff
@@ -11,40 +11,40 @@
...
+    """Represents an open session with a Jupyter Lab.
+
+    This holds the information a client needs to talk to the Lab in order to
+    execute code.
+    """
 
     session_id: str
     kernel_id: str
-    websocket: Optional[ClientWebSocketResponse]
+    websocket: ClientWebSocketResponse
```
athornton, line 47: Ah!  AsyncMock to the rescue!

src/mobu/jupyterclient.py
```diff
@@ -62,6 +62,15 @@ class JupyterClientSession:
         The session to wrap.
     token : `str`
         The token to send.
+
```
athornton, line 65: Thank you for the comment.  This makes the intent and limitations much clearer.

src/mobu/jupyterclient.py
```diff
@@ -158,98 +165,130 @@ async def lab_login(self) -> None:
...
-    async def is_lab_running(self) -> bool:
-        hub_url = self.jupyter_url + "hub"
-        spawn_url = self.jupyter_url + "hub/spawn"
-        spawn_pending_url = (
-            self.jupyter_url + f"hub/spawn-pending/{self.user.username}"
-        )
-        lab_url = self.jupyter_url + f"user/{self.user.username}/lab"
-
-        async with self.session.get(hub_url) as r:
+    async def lab_is_stopped(self) -> bool:
```
athornton, line 168: This is really bletcherous.  I get why it's necessary.  I will go ask on the gitter whether this is already an extant API, and if the answer is "er, no" then maybe creating that and submitting the PR should be a priority for us, because surely we're not the only people who need it.

src/mobu/jupyterclient.py
```diff
@@ -158,98 +165,130 @@ async def lab_login(self) -> None:
...
 
         # Wait for the lab to actually go away.  If we don't do this, we may
         # try to create a new lab while the old one is still shutting down.
-        count = 0
-        while await self.is_lab_running() and count < 10:
-            self.log.info(f"Waiting for lab deletion ({count}s elapsed)")
-            await asyncio.sleep(1)
-            count += 1
-        if await self.is_lab_running():
+        max_poll_secs = 20
```
athornton, line 280: I would probably make this 30, just because I'm a little twitchy about what's going to happen on underpowered T&S teststands.

src/mobu/jupyterclient.py
```diff
@@ -356,5 +384,5 @@ async def run_python(self, session: JupyterLabSession, code: str) -> str:
                 else:
                     raise NotebookException(f"Result status is {status}")
 
-    async def _raise_error(self, msg: str, r: ClientResponse) -> None:
+    async def _raise_error(self, msg: str, r: ClientResponse) -> NoReturn:
```
athornton, line 387: TIL about NoReturn

athornton (approved):
I like it.  Encapsulating the nasty lab-state-determination, especially.

src/mobu/business/jupyterloginloop.py
rra: Yeah, I moved this from JupyterClient to the business so that we get logging and timing.

src/mobu/jupyterclient.py
rra: Thank you! Yeah, it seems obvious there should be an API for this.

src/mobu/jupyterclient.py
rra: I want to figure out how to make more of these waits interruptable and then it will be less worrisome to increase them. FWIW, on the IDF the deletion usually happens within 5s, but agreed that underpowered teststands could be a problem.

PR #81, opened 2021-08-18
[DM-31406] Prepare mobu for running the tutorial notebook

rra:
- Add support for setting the working directory when running a notebook
- Add support for getting the node on which a notebook is running and reporting it in alerts
- Catch and translate the poor aiohttp exception from a closed WebSocket
- Improve the way we determine what notebooks to run
- Report the cell of the notebook that was being executed when an alert fires
- Simplify the timing data that's gathered and reported
- Add divider after unknown exception reports to make them easier to read

### DIFF
--- a/src/mobu/business/notebookrunner.py
+++ b/src/mobu/business/notebookrunner.py
@@ -98,14 +98,12 @@ class NotebookRunner(JupyterPythonLoop):
         for count in range(self.config.max_executions):
             self.next_notebook()
             assert self.notebook
-            self.logger.info(f"Starting notebook: {self.notebook.name}")
-            cells = self.read_notebook(self.notebook)
 
             iteration = f"{count + 1}/{self.config.max_executions}"
             msg = f"Notebook {self.notebook.name} iteration {iteration}"
             self.logger.info(msg)
 
-            for cell in cells:
+            for cell in self.read_notebook(self.notebook):
                 self.running_code = "".join(cell["source"])
                 await self.execute_cell(session, self.running_code)
                 self.running_code = None
--- a/src/mobu/business/notebookrunner.py
+++ b/src/mobu/business/notebookrunner.py
@@ -62,6 +62,12 @@ class NotebookRunner(JupyterPythonLoop):
             self._repo = git.Repo.clone_from(url, path, branch=branch)
 
     async def execute_code(self, session: JupyterLabSession) -> None:
+        if self.config.working_directory:
+            code = f'import os; os.chdir("{self.config.working_directory}")'
+            with self.timings.start("execute_setup", {"code": code}) as sw:
+                reply = await self._client.run_python(session, code)
+                sw.annotation["result"] = reply
+
         for count in range(self.config.max_executions):
             self._next_notebook()
             assert self.notebook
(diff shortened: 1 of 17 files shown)


rra:
Now tested on data.lsst.cloud and the tutorial notebook runs successfully.

src/mobu/business/notebookrunner.py
```diff
@@ -61,73 +60,94 @@ def clone_repo(self) -> None:
...
+    def read_notebook(self, notebook: Path) -> List[Dict[str, Any]]:
+        with self.timings.start("read_notebook", {"notebook": notebook.name}):
+            try:
+                notebook_text = notebook.read_text()
+                cells = json.loads(notebook_text)["cells"]
+            except Exception as e:
+                msg = f"Invalid notebook {notebook.name}: {str(e)}"
+                raise NotebookRepositoryError(msg)
+
+        # Add cell numbers to all the cells, which we'll use to name timing
```
athornton, line 91: I love this.

src/mobu/business/notebookrunner.py
```diff
@@ -61,73 +60,94 @@ def clone_repo(self) -> None:
...
+        with self.timings.start("find_notebooks"):
+            notebooks = [
+                p
+                for p in Path(self._repo_dir.name).iterdir()
+                if p.suffix == ".ipynb"
+            ]
+            if not notebooks:
+                msg = "No notebooks found in {self._repo_dir.name}"
+                raise NotebookRepositoryError(msg)
+            SystemRandom().shuffle(notebooks)
```
athornton, line 73: I think we might want to make the shuffle configurable.  I can imagine some sort of use-case where we wanted to ensure the order in which we were running things.

src/mobu/monkey.py
```diff
@@ -113,13 +113,16 @@ async def _runner(self) -> None:
                 if self.state == MonkeyState.RUNNING:
                     self.state = MonkeyState.ERROR
             if run:
+                self.log.warning("Restarting failed monkey after 60s")
```
athornton, line 116: should 60 be a parameter?

tests/business/notebookrunner_test.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Wait until we've finished at least one loop and check the results.
+    data = await wait_for_business(client, "testuser1")
+    assert data == {
+        "name": "testuser1",
+        "business": {
+            "failure_count": 0,
+            "name": "NotebookRunner",
+            "notebook": "test-notebook.ipynb",
+            "success_count": 1,
```
athornton, line 72: If this is correct, then the comment on line 64 (and 140) should lose "at least," I think.

athornton (approved):
Minor comments, but overall looks great.

src/mobu/business/notebookrunner.py
rra: I agree, but I think I'll defer that until we have a use case, since I suspect it will need something more complicated than turning off shuffling (providing a list of notebooks to run in order, for instance).

src/mobu/monkey.py
rra: Meh, we haven't needed to tune it yet. We can add a parameter pretty easily when we run into a case where we want to change it.

tests/business/notebookrunner_test.py
rra: Good call, fixed.

PR #82, opened 2021-08-20
[DM-31454] Support TAP testing

rra:
- Refactor QueryMonkey to bring it in line with the work done in other parts of mobu
- Rename it to TAPQueryRunner for parallelism with NotebookRunner
- Update the query templates to match the ones currently used by querymonkey
- Fix some tricky log doubling issues with pyvo by disabling log propagation
- Update dependencies


src/mobu/jupyterclient.py
```diff
@@ -314,7 +314,7 @@ async def create_labsession(
         try:
             websocket = await self.session.ws_connect(channels_url)
         except WSServerHandshakeError as e:
-            raise JupyterError.from_exception(self.user.username, e)
+            raise JupyterError.from_exception(self.user.username, e) from None
```
athornton, line 317: This is novel syntax to me.

athornton: https://docs.python.org/3/library/exceptions.html cool

athornton (approved):
Looks good.  Didn't know about raise from.

cbanek (approved):

PR #84, opened 2021-08-24
[DM-31504] Handle failure to delete a lab

rra:
If a lab fails to delete in the timeout period, log some additional
information about the return status from the hub.  If the failure
happens during startup, continue on to try to spawn the lab anyway
in the hope that JupyterHub will be able to clean up inconsistent
state.


athornton (approved):
LGTM

PR #87, opened 2021-08-31
[DM-31490] Overhaul how Jupyter images are selected

rra:
Rather than configuring mobu with the raw values to post to the
spawner form, support asking cachemachine for an image satisfying
a particular quality (recommended and latest weekly being the main
supported ones right now), and then determine the correct thing to
send to the spawner page from that.

Use this infrastructure to record the image name in error reports
and in metadata about a running monkey business.


athornton (approved):
LGTM.  Interesting that we actually *want* to couple our services here, but, yeah, I think we do.

PR #88, opened 2021-09-01
[DM-31581] Trim errors and code to 3000 characters

rra:
Slack has a limit on block sizes of 3000 characters, but sometimes
errors can be longer than that.  Trim errors and code from the top
(since the bottom is usually more interesting) to get it below
Slack's limit so that we don't have missing alerts.


athornton (approved):
LGTM

PR #91, opened 2021-09-15
Accomodate module change for JL3 move

athornton:


rra:
I have a mild preference for using a somewhat more succinct name for this module since it includes helpers that users may want to import and run directly. Let's talk about this at co-work.

athornton:
Module name is now lsst.rsp.  Closing this one unmerged.

PR #101, opened 2021-11-17
tickets/DM-32511: Add spawntimer business

athornton:
This adds slightly more granularity during Lab spawn and preserves the event log messages generated.  I hope to use this to correlate K8s messages with slow-to-spawn-because-(I-think-)NFS-mounts-are-slow-sometimes.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -122,6 +122,7 @@ certifi==2021.10.8 \
     --hash=sha256:d62a0163eb4c2344ac042ab2bdf75399a71a2d8c7d47eac2e2ee91b9d6339569
     # via
     #   -c requirements/main.txt
+    #   httpcore
     #   httpx
 cfgv==3.3.1 \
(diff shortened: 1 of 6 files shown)


rra (commented):
We should just add the timing granularity directly to the normal class rather than adding a new business, I think. I'm planning on getting rid of annotations completely to try to reduce the memory pain, so we probably need a broader conversation about why you wanted to add them and how to capture that data going forward.

You'll want to start from #99 instead of master, I think, because otherwise there will be a lot of merging pain.

athornton:
OK, I can do that.  Maybe not today.  I can do without the annotations as long as I have somewhere I can stash the actual event messages that come out of the watcher.

rra:
The thought was that they were in the log of the monkey, so you can get them with `/log`. That may not be in the best format, though. (I'm not sure whether making the log JSON would be better or worse; it makes it harder for humans but easier for machines.)

PR #145, opened 2022-06-09
[DM-34613] Make UID optional for users

rra:
In some environments, Gafaelfawr will be able to automatically
assign UIDs.  Make providing a UID or a UID spec optional, so that
we can just omit UIDs in those environments.


src/mobu/models/user.py
```diff
@@ -16,7 +16,12 @@ class User(BaseModel):
 
     username: str = Field(..., title="Username", example="testuser")
 
-    uidnumber: int = Field(..., title="Numeric UID", example=60001)
+    uidnumber: Optional[int] = Field(
+        None,
+        title="Numeric UID",
+        description="If omitted, ask Gafaelfawr to assign a UID",
```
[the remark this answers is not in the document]
jonathansick: (the same text appears in multiple places)

src/mobu/main.py
```diff
@@ -47,12 +47,14 @@
 app.include_router(internal_router)
 app.include_router(external_router, prefix=f"/{config.name}")
 
+# Add middleware.
+app.add_middleware(XForwardedMiddleware)
```
jonathansick, line 51: Curious to learn more about this. Is this a performance concern specific to mobu or do we need to think about this in other apps too?

Is the concern that app set up is taking longer, or that the test requests are actually running literally through multiple copies of the middleware? Can we do something about this in our middleware code?

jonathansick (approved):

src/mobu/main.py
rra: It's the latter; `add_middleware` doesn't check if the middleware is already added and will keep adding another copy to the middleware stack. This appears to be by design. There isn't anything we can do in our middleware code about it that I can see; we can do an early exit if we've already run, but the cost of the overhead is from all the layered calls, not from the (small) amount of work the middleware does.

I think conceptually this was my mistake in ever initializing middleware from a startup hook. It's best thought of like a router that should be included during app creation.

PR #149, opened 2022-06-16
[DM-35246] Add more query annotations to TAP queries

rra:
Add a static "monkey" string to each SELECT clause in the TAP
queries (previously, the histograms were missing one).  Add a
second static string specifying the query template to each query.

This makes it easier to generate statistics on the qserv side.


cbanek (approved):
LGTM!

PR #169, opened 2022-08-12
[DM-35881] Add GID support

rra:
Allow explicitly specifying the GID, either per user or with a
spec.  Default to setting the GID to the same as the UID, which was
the behavior previously forced in nublado2, if the UID was specified.
If it was not specified, leave the behavior to Gafaelfawr, but
document that synthetic user private groups need to be enabled.


src/mobu/models/user.py
```diff
@@ -26,6 +26,18 @@ class User(BaseModel):
         example=60001,
     )
 
+    gidnumber: Optional[int] = Field(
```
jonathansick, line 29: In what circumstance does mobu need to control the `uid` and `gid` of its bot users? I deprecated that in noteburst, so is this a hold-over for certain environments using certain versions of gafaelfawr?

jonathansick (approved):

src/mobu/models/user.py
rra: mobu will need to control the UID and GID in environments where we don't assign UIDs and GIDs and where the users that mobu is generating don't appear in the local identity management system (such as LDAP), since Gafaelfawr doesn't know how to assign UIDs and GIDs that don't conflict in such environments. (Also, the local administrator may have to pre-create home directories or the like, since moneypenny may not be in use there.)

mobu will also need to assign UIDs and GIDs in GitHub environments, since similarly Gafaelfawr doesn't know what to do there, although there the problem is that what to do is unknowable: there's basically no alternative to just picking something and hoping we don't have a collision, I think.

I'm not sure what environments noteburst will need to be deployed in. If it's only the IDF, you won't need to care, but for the T&S deployments Gafaelfawr cannot allocate UIDs and GIDs since they come from GitHub.

PR #199, opened 2023-02-03
DM-37758: Replace Cachemachine with JupyterLab Controller

athornton:


src/mobu/jupyterclient.py
```diff
@@ -501,10 +506,12 @@ def _remove_ansi_escapes(string: str) -> str:
...
+    def _build_jupyter_spawn_form(
+        self, image: ControllerImage
+    ) -> Dict[str, List[str]]:
         """Construct the form to submit to the JupyterHub login page."""
         return {
-            "image_list": str(image),
-            "image_dropdown": "use_image_from_dropdown",
-            "size": self.config.image_size,
+            "image_list": [image.path],
+            "image_dropdown": [image.path],  # Not used
```
rra, line 515: Does this have to be sent? It would be nice if it could be omitted since it's not used.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+    This isn't actually all that useful here, but since these models are
+    copied from jupyterlabcontroller, which *does* make use of these features,
+    it's easier than messing with the models.
+    """
+
+    class Config:
+        """Pydantic configuration."""
+
+        alias_generator = to_camel_case
+        allow_population_by_field_name = True
```
rra, line 51: I don't think you end up needing any of this code since you use dashify instead.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+    will produce auto() values in lower case.  This is exactly StrEnum from
+    Python 3.11, except for the validation step."""
 
-    reference: str = Field(
+    def _generate_next_value_(  # type: ignore
+        name, start, count, last_values
+    ) -> str:
+        if name != name.upper():
+            raise RuntimeError("Enum names must be entirely upper-case")
+        return name.lower()
```
rra, line 86: I don't understand these changes. They seem like a ton of unnecessary complexity. Why can't we just use simple enums?

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
-        description="May be null if the digest isn't known",
+        name="digest",
         example=(
-            "sha256:419c4b7e14603711b25fa9e0569460a753c4b2449fe275bb5f89743b"
-            "01794a30"
+            "sha256:e693782192ecef4f7846ad2b21"
+            "b1574682e700747f94c5a256b5731331a2eec2"
         ),
+        title="unique digest of image contents",
+    )
```
rra, line 111: This should be optional since we don't care about it, I think. Then when the image is manually constructed for the case where we explicitly provide an image, we don't need to pass in a dummy string and can just let it default to `None`.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+        alias_generator = dashify
+        allow_population_by_field_name = True
+
+
+class JupyterImageClass(SpawnerEnum):
+    """Possible ways of selecting an image."""
+
+    RECOMMENDED = auto()
+    LATEST_WEEKLY = auto()
+    BY_REFERENCE = auto()
```
rra, line 158:
```suggestion
class JupyterImageClass(Enum):
    """Possible ways of selecting an image."""

    RECOMMENDED = "recommended"
    LATEST_WEEKLY = "latest-weekly"
    BY_REFERENCE = "by-reference"
```

tests/autostart_test.py
```diff
@@ -129,11 +129,12 @@ async def test_autostart(client: AsyncClient) -> None:
                     "failure_count": 0,
                     "image": {
                         "digest": ANY,
-                        "name": "Weekly 2021_35",
-                        "reference": (
-                            "registry.hub.docker.com/lsstsqre/sciplat-lab"
-                            ":w_2021_35"
+                        "name": "Weekly 2023_05",
+                        "path": (
+                            "docker.io/lsstsqre/sciplat-lab" ":w_2023_05"
```
rra, line 134: These strings can be combined now.

tests/autostart_test.py
```diff
@@ -155,11 +156,12 @@ async def test_autostart(client: AsyncClient) -> None:
...
```
rra, line 161: Strings can be combined now.

tests/business/jupyterpythonloop_test.py
```diff
@@ -203,8 +203,7 @@ async def test_long_error(
                 "jupyter": {
                     "image_class": "by-reference",
                     "image_reference": (
-                        "registry.hub.docker.com/lsstsqre/sciplat-lab"
-                        ":d_2021_08_30"
+                        "docker.io/lsstsqre/sciplat-lab" ":d_2023_02_03"
```
rra, line 206: Strings can be combined now.

src/mobu/jupyterclient.py
```diff
@@ -295,17 +296,21 @@ async def is_lab_stopped(self, final: bool = False) -> bool:
...
+            image = await self.controller.get_recommended()
         elif self.config.image_class == JupyterImageClass.LATEST_WEEKLY:
-            image = await self.cachemachine.get_latest_weekly()
+            image = await self.controller.get_latest_weekly()
         else:
             assert self.config.image_reference
-            image = JupyterImage.from_reference(self.config.image_reference)
+            ref = self.config.image_reference
+            image = ControllerImage(
+                path=ref, name=ref, digest="unknown", tags={}
```
rra, line 311: `digest` and `tags` should be optional in the model so that you can just omit them here.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+class ControllerImages(CamelCaseModel):
+    recommended: Optional[ControllerImage] = None
+    latest_weekly: Optional[ControllerImage] = None
+    latest_daily: Optional[ControllerImage] = None
+    latest_release: Optional[ControllerImage] = None
+    all: List[ControllerImage] = Field(default_factory=list)
+
+    class Config:
+        alias_generator = dashify
+        allow_population_by_field_name = True
```
rra, line 150: This is defined as a `CamelCaseModel` but then you override everything `CamelCaseModel` does, so this should just be `BaseModel`.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+    def _generate_next_value_(  # type: ignore
+        name, start, count, last_values
+    ) -> str:
+        if name != name.upper():
+            raise RuntimeError("Enum names must be entirely upper-case")
+        return name.lower()
+
+
+class PartialImage(CamelCaseModel):
+    path: str = Field(
```
rra, line 90: Not sure why this isn't merged with `ContainerImage`. I don't think mobu ever cares about `PartialImage`.

src/mobu/models/jupyter.py
```diff
@@ -1,62 +1,161 @@
...
+        title="unique digest of image contents",
+    )
+
+
+class ControllerImage(PartialImage):
+    tags: Dict[str, str] = Field(
+        ...,
+        name="tags",
+        title="Map between tag and its display name",
+    )
```
rra, line 119: Should be optional so that you don't need to supply it when constructing the image by reference.

src/mobu/jupyterclient.py
athornton: I think I can modify the models on the receiving end to make those optional.  At least one of list or dropdown needs to be sent, but I can certainly send dropdown as null.

src/mobu/models/jupyter.py
athornton: ControllerImages is a CamelCase model and I wanted to minimize the changes when I forklifted the model from jupyterlabcontroller.

That is a question I've got, actually: Ideally models shared between applications (e.g. as producer and consumer, which is what we're doing here) should go in some common library they both reference.  What's a good pattern for that?

src/mobu/models/jupyter.py
athornton: Same as above: I tried to minimize the changes when pulling this over from JupyterLab Controller.  We *do* care about the actual string values there.  I think the right way to go may be to go to Python 3.11 and use its new StrEnum class directly, though.  That doesn't get us the keys-are-uppercase validator, but, meh.

src/mobu/models/jupyter.py
athornton: This comes back to the shared-class thing.  It makes sense for it to be a CamelCaseModel in the JupyterLab Controller, because we instantiate it directly from a CamelCase YAML file.  It doesn't really make sense here, but I also feel like it should be as nearly as possible the same model in each place, or we should have a repository or set of repositories that hold models that are used as things we're passing around over the network.

src/mobu/models/jupyter.py
athornton: One more "well, I wanted to leave the classes from the controller intact."

src/mobu/models/jupyter.py
jonathansick: I've been pondering the same thing for Pydantic-based Avro for Roundtable Kafka producers and consumers. I was thinking about putting them in Safir?

rra:
This was done using the bot API instead of talking to the lab controller directly in #227.

PR #235, opened 2023-04-27
DM-38425: Explicitly log out of the hub during startup

rra:
We encountered some sort of corrupt user state where all logins failed due to an internal error in the JupyterHub OpenID Connect server saying there was a mismatch of the client ID. Try explicitly logging out during startup and after an error to see if that will correct that problem.


rra:
This didn't resolve the problem and shouldn't be necessary.

PR #302, opened 2023-09-28
Add option to use query lists in values config in TAPQueryRunner (Options)

stvoutsin:
Changes introduce a way of defining queries to be executed by a Mobu TAP Test run in the configuration. Specifically, with the changes an RSP maintainer can specify the queries in the following way (or use the existing query_set option to use a local filepath / template):

```
config:
  autostart:
    - name: "tap"
      count: 1
      users:
	- username: "bot-mobu-tap"
	  uidnumber: 74775
	  gidnumber: 74775
      scopes: ["read:tap"]
      business:
	type: "TAPQueryRunner"
	restart: true
	options:
	  sync: true
	  queries: 
	    - "SELECT TOP 10 * FROM TAP_SCHEMA.schemas"
	    - "SELECT TOP 10 * FROM MYDB.MyTable"
```

Changes include a couple more minor changes where my formatter thought the length of a line was too long, this can be excluded from the PR if you think they don't need to be included. Also one more test added, thought it could use a few more to improve coverage.

Let me know if you'd like any of the changes modified / removed, as this was a first attempt to get this working, but I'm not sure if it fits with your initial designs and/or if the naming of the methods/classes match what the code is doing.

### DIFF
--- a/src/mobu/models/business/tapqueryrunner.py
+++ b/src/mobu/models/business/tapqueryrunner.py
@@ -34,13 +34,18 @@ class TAPQueryRunnerOptions(BusinessOptions):
         example=True,
     )
 
+    queries: list | None = Field(
+        None,
+        title="Which query list to use for the TapQueryRunner",
+        description="List of queries to be run instead of a query_set",
+        example=True,
+    )
+
 
 class TAPQueryRunnerConfig(BusinessConfig):
     """Configuration specialization for TAPQueryRunner."""
 
-    type: Literal["TAPQueryRunner"] = Field(
-        ..., title="Type of business to run"
-    )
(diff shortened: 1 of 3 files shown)


rra:
I have finally merged a version of this in https://github.com/lsst-sqre/mobu/pull/312. Apologies for the long delay in taking a look.

The protocol approach here was intriguing and I will keep that technique in mind, but since mobu had already gone down the path of subclassing for behavior, I took a somewhat less backward-compatible approach and renamed `TAPQueryRunner` to `TAPQuerySetRunner` so that I could add your new code as `TAPQueryRunner` and move shared code into an abstract base class. This avoids the optional configuration problem and makes what the business is doing unambiguous from its configuration.

Thank you very much for the PR!

PR #339, opened 2024-03-12
tickets/DM:43203: add Git-LFS business

athornton:
Add Git-LFS business.


src/mobu/services/business/notebookrunner.py
```diff
@@ -68,18 +68,18 @@ def annotations(self, cell_id: str | None = None) -> dict[str, str]:
         return result
 
     async def startup(self) -> None:
-        if not self._repo:
-            self.clone_repo()
+        if not any(Path(self._repo_dir.name).rglob("*")):
+            # https://stackoverflow.com/questions/25675352/
+            await self.clone_repo()
```
rra, line 73: This seems unnecessarily confusing, so much so that it requires a comment pointing to a StackOverflow answer. Why not stick with the existing pattern where the Git storage is created during `startup` if it doesn't already exist, and use that as the trigger to clone the repository?

src/mobu/storage/git.py
```diff
@@ -0,0 +1,73 @@
...
+from pathlib import Path
+
+from structlog.stdlib import BoundLogger
+
+from ..models.user import AuthenticatedUser
+from .process import Process
+
+
+class Git:
+    """A very basic async Git client based on asyncio.subprocess."""
```
rra, line 14: Missing documentation of constructor parameters.

src/mobu/storage/git.py
```diff
@@ -0,0 +1,73 @@
...
+                gidnumber=60001,
+                scopes=[],
+                token="dummy token",
+            )
+        self._logger = logger
+        self._repo = repo
+        self._process = Process(user=self._user, logger=self._logger)
+
+    def set_repo(self, repo: Path) -> None:
+        self._repo = repo
```
rra, line 37: At first glance, the `set_repo` pattern feels like unnecessary state manipulation. Why not create a new `Git` object for every repo path? It looks like a lightweight object.

src/mobu/storage/git.py
```diff
@@ -0,0 +1,73 @@
...
+    ) -> None:
+        self._user = user
+        if self._user is None:
+            self._user = AuthenticatedUser(
+                username="gituser",
+                uidnumber=60001,
+                gidnumber=60001,
+                scopes=[],
+                token="dummy token",
+            )
```
rra, line 31: All of this user stuff appears to be used only to add the username to the `AsyncProcessError` exception. I think you can delete all of it; the monkey already handles adding the user when reporting failures.

src/mobu/storage/git.py
```diff
@@ -0,0 +1,73 @@
...
+        self._logger = logger
+        self._repo = repo
+        self._process = Process(user=self._user, logger=self._logger)
+
+    def set_repo(self, repo: Path) -> None:
+        self._repo = repo
+
+    async def git(self, *args: str) -> None:
```
rra, line 39: docstrings needed for all of these methods.

src/mobu/storage/process.py
```diff
@@ -0,0 +1,57 @@
...
+from shlex import join
+
+from structlog.stdlib import BoundLogger
+
+from ..exceptions import AsyncioProcessError
+from ..models.user import AuthenticatedUser
+
+
+class Process:
+    """A thin wrapper around asyncio.subprocess.create_process_exec."""
```
rra, line 14: I don't think this is sufficiently generic to warrant being a separate class as opposed to a method on `Git`. It doesn't have any state that isn't already in the `Git` object state.

src/mobu/storage/process.py
```diff
@@ -0,0 +1,57 @@
...
+    ) -> None:
+        self._user = user
+        self._logger = logger
+
+    async def exec(
+        self,
+        cmd: str,
+        *args: str,
+        cwd: Path | None = None,
+    ) -> asyncio.subprocess.Process:
```
rra, line 30: The return value here is not currently being used, so I would start with the simplest interface that will work, don't return a value, and we can always make it more complicated later if we have to.

src/mobu/storage/process.py
```diff
@@ -0,0 +1,57 @@
...
+        msg += f" exited with rc={proc.returncode}"  # which should be 0
+        if proc.stdout is not None:
+            msg += f'; stdout="{stdout.decode()}"'
+        if proc.stderr is not None:
+            msg += f'; stderr="{stderr.decode()}"'
+        if proc.returncode != 0:
+            raise AsyncioProcessError(msg, user=username)
+
+        if self._logger:
+            self._logger.debug(msg)
```
rra, line 56: All this structure should be preserved in both the exception and in the logger invocation rather than being flattened into a string. I think the place to do the flattening would be in a `__str__` method on the exception.

src/mobu/models/business/gitlfs.py
```diff
@@ -0,0 +1,36 @@
...
+        title="LFS read URL for Git-LFS enabled repo",
+    )
+    lfs_write_url: str = Field(
+        "https://git-lfs-rw.lsst.cloud/mobu/git-lfs-test",
+        title="LFS write URL for Git-LFS enabled repo",
+    )
+    skip_lfs: bool = Field(
+        False,
+        title="Skip LFS operations (for testing only)",
+    )
```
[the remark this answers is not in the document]
athornton: Hm.  If I refactor the business to make the LFS-specific steps their own method(s), yeah.

src/mobu/services/business/gitlfs.py
```diff
@@ -0,0 +1,164 @@
...
+        self.logger.info("Running Git-LFS check:")
+        with self.timings.start("execute git-lfs check") as sw:
+            try:
+                await self._git_lfs_check()
+            except Exception as e:
+                raise GitError(e, user=self.user.username) from e
+            elapsed = sw.elapsed.total_seconds()
+        self.logger.info(f"Git-LFS check finished after {elapsed} seconds")
+
+    async def _git_lfs_check(self) -> None:
```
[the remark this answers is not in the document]
athornton: The fact that we push back to the upstream makes it slightly tougher, and I think I am confused about how `startup()` works.  But in general, yeah, I broke it up.

src/mobu/services/business/gitlfs.py
```diff
@@ -0,0 +1,164 @@
...
+        if not self._skip_lfs:  # Only EVER skip LFS for unit testing.
+            with self.timings.start("add LFS configuration to commit"):
+                await self._git.config(
+                    "--local", "lfs.url", self._lfs_write_url
+                )
+                await self._git.config("--local", "lfs.locksverify", "false")
+                await self._git.config(
+                    "--local", "lfs.repositoryformatversion", "0"
+                )
+                await self._git.config("--local", "lfs.access", "basic")
```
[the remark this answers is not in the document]
athornton: I'm trying this now, but `GIT_CONFIG_GLOBAL` doesn't seem to be respected.  You can see it in the Slack logs for #roundtable-dev, where the variable is clearly set but it doesn't know `user.email` and `user.name`.

src/mobu/services/business/notebookrunner.py
athornton: At least conceptually, sure.  I didn't go back to the main branch, but we only set self._repo_dir in startup, and then only if the value is None.  Then we remove it in shutdown (and set it back to None, not that it matters, since the Business object is about to go away).

src/mobu/storage/git.py
athornton: Done.  The short answer is, it's a fair bit of ceremony each time.  But it wasn't too bad.

src/mobu/services/business/gitlfs.py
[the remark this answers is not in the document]
athornton: And then I put it back together, kinda.  I mean, it's different methods, but I'm destroying the entire working directory after each time through the loop.  I'm initializing the origin repo with a file copy now, but there's a lot less bookkeeping if we just start from scratch each time.

src/mobu/services/business/gitlfs.py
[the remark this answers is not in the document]
athornton: Turned out to be a too-old git shipped with buster; going to bookworm fixed that.

athornton:
I don't understand about the conflicts.  I thought it was a straightforward dependency update, but apparently not?

athornton:
...it was overzealous squashing, I think.  Anyway, I got it happy with the merge conflicts and then updated again and it's fine.

src/mobu/storage/git.py
```diff
@@ -0,0 +1,255 @@
...
+                cwd=cwd,
+                env=sub_env,
+            )
+        if self._logger:
+            msg = (
+                f"'{cmd_and_args}' exited with rc={proc.returncode};"
+                f" stdout='{stdout_text}'; stderr='{stderr_text}'"
+                f" working_directory='{cwd!s}'; env='{sub_env}'"
+            )
+            self._logger.debug(msg)
```
rra, line 107: All those key/value pairs should be key/value arguments to `debug`, not collapsed into a string. We're using structlog exactly so that we can preserve structure like this as JSON in the log message.

tests/business/gitlfs_test.py
```diff
@@ -0,0 +1,111 @@
...
+            "_check_uuid_pointer",
+            side_effect=verify_uuid_contents,
+            autospec=True,
+        ):
+            with patch.object(
+                GitLFSBusiness,
+                "_add_git_lfs_data",
+                side_effect=no_git_lfs_data,
+                autospec=True,
+            ):
```
rra, line 43: If you do these as fixtures (you can define test-specific fixtures directly in the test file just like you do in `conftest.py`), your code won't have to drift as far to the right.

rra (commented):
Thanks, this looks pretty good, just a few remaining notes. I'm still leery of doing this much work and this many forks for ever iteration of the test, but this is fine for now and we can always look at it later if need be.

tests/business/gitlfs_test.py
rra: Or, probably better, a single fixture, since they don't really make sense by themselves. Basically just do the same series of `with` statements but then run `yield` in the body, and you have a yield fixture that yields `None` that you can then reference in the test as `mock_git_lfs: None` or whatever as an argument.

src/mobu/storage/git.py
```diff
@@ -0,0 +1,255 @@
...
+        if proc.stderr is not None:
+            stderr_text = stderr.decode()
+        if proc.returncode != 0:
+            raise SubprocessError(
+                f"Subprocess '{cmd_and_args}' failed",
+                returncode=proc.returncode,
+                stdout=stdout_text,
+                stderr=stderr_text,
+                cwd=cwd,
+                env=sub_env,
```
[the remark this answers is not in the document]
athornton: Oh, OK, so just copy anything I _need_ out of the real environment, rather than updating a whole copy.

src/mobu/services/business/gitlfs.py
```diff
@@ -0,0 +1,248 @@
...
+        credfile.touch()
+        credfile.chmod(0o700)
+        w_url = urlparse(self._lfs_write_url)
+        await git.config(
+            "--local",
+            f"credential.{w_url.scheme}://{w_url.netloc}.helper",
+            f"store --file {credfile!s}",
+        )
+        creds = f"{w_url.scheme}://gituser:{self.user.token}@{w_url.netloc}\n"
+        credfile.write_text(creds)
```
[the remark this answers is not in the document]
athornton: Yeah, it's probably dumb.  What I'm doing is having the credfile containing a secret extant for the minimum possible length of time.

athornton:
The thing about the number of forks is that to see what it's saving us we would have to look through GitPython to see how many of _its_ calls end up doing a fork to run `git` under the hood.  But also, yeah, I feel like it's not 1994 anymore and the whole thing runs, including the push to real git-lfs, in less than three seconds.

src/mobu/services/business/notebookrunner.py
```diff
@@ -68,28 +69,33 @@ def annotations(self, cell_id: str | None = None) -> dict[str, str]:
...
-            notebooks = [
-                p
-                for p in Path(self._repo_dir.name).iterdir()
-                if p.suffix == ".ipynb"
-            ]
+            notebooks: list[Path] = []
+            if self._repo_dir is not None:
+                notebooks = [
+                    p for p in self._repo_dir.iterdir() if p.suffix == ".ipynb"
+                ]
```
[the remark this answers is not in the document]
athornton: In any case it's not going to happen, because that git client creation has a specific repo set during creation, but mypy doesn't know that.  I'll make it a ValueError.

src/mobu/services/business/gitlfs.py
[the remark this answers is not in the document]
rra: Oh, okay. I wouldn't bother; the token is pretty low-value anyway.

src/mobu/storage/git.py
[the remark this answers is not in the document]
rra: Well, you can do that, but that wasn't what I was saying. I was just saying that when you construct the exception, which will report its data to Slack, only include the environment variables that you have explicitly added for Git, not all the ones that come from mobu's environment. You can pass them all down to subprocess but only the relevant additional ones to the exception (and to logging).

src/mobu/services/business/gitlfs.py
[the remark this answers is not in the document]
athornton: I've put all the actual file manipulation in the same place, and collapsed the touch and chmod.

rra:
I'm not talking about minimizing forks compared to what GitPython was doing, but to do the setup of the upstream and one of the downstream repositories once so that you don't have to run all the git commands for them every time, like I mentioned in the previous review. But this is simpler, so sure, let's go with this for right now.

src/mobu/storage/git.py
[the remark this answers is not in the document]
athornton: Well, I think the subprocess only *really* needs PATH.  HOME and LANG might be nice too.

athornton:
Oh.  Yeah, the effort needed to restore origin and Downstream #1 to their initial state to let us re-run the test is maybe less than rerunning all the git commands every time, but I find it much harder to think about than just wiping out the directory and starting afresh.

src/mobu/exceptions.py
```diff
@@ -37,6 +39,8 @@
     "MobuSlackWebException",
     "MonkeyNotFoundError",
     "TAPClientError",
+    "SubprocessError",
+    "ComparisonError",
```
rra, line 44: Sort alphabetically.

rra (approved):
Needs a change log entry (create with `scriv create` and then edit the resulting file) and one minor nit, otherwise should be good to go.

PR #340, opened 2024-03-21
DM-43423: Fix XSRF cookie and header handling for JupyterLab

rra:
Previously, mobu created a fake XSRF cookie and header and sent them blindly to JupyterLab. JupyterLab now expects the cookie to be in a specific format and is rejecting our fake one.

Instead, properly implement the same protocol as the JavaScript client: retrieve the _xsrf cookie from the response from the lab and lift it into the X-XSRFToken header. Raise an exception if the lab does not send an _xsrf cookie so that we'll catch further changes more quickly.

This is complicated by the fact that the hub and lab use separate XSRF tokens restricted to different paths, so we need to track both tokens and send the appropriate header based on which API we're talking to.


athornton (approved):
I'm not sure I'm 100% clear on what's going on, but the gist is, we had been just making something up for an XSRF token, and now we're having to do the more-right thing of checking the cookie we get and putting the token in there into our HTTP headers?

I don't see anything to object to, especially if it works.

PR #347, opened 2024-05-19
DM-44397: Run all notebooks in repo directory 

fajpunk:
* NotebookRunner business now runs all notebooks in a repo by default
* Except in the `mobu_exclude` or its descendants 
* Add `exclude_dirs` option to NotebookRunner business to list
  additional directories in which notebooks will NOT be run
* Phalanx mobu docs updated [here](https://github.com/lsst-sqre/phalanx/pull/3346).

Before this goes out, we'll need to update the [tutorial notebooks](https://github.com/rubin-dp0/tutorial-notebooks.git) and the [system-test](https://github.com/lsst-sqre/system-test/pull/84) notebook repos to put all of the subdir notebooks in the `mobu_exclude` dir.


src/mobu/constants.py
```diff
@@ -5,13 +5,17 @@
...
     "NOTEBOOK_REPO_URL",
     "NOTEBOOK_REPO_BRANCH",
     "TOKEN_LIFETIME",
     "USERNAME_REGEX",
     "WEBSOCKET_OPEN_TIMEOUT",
 ]
 
+NOTEBOOK_ALWAYS_EXCLUDE_DIRS = {"mobu_exclude"}
+"""NotebookRunner always excludes notebooks in these dirs."""
+
```
rra, line 18: This feels a little magical and hard-coded to me. I think I would lean towards not excluding any directories by default and making people add an excluded directory if they want one. But maybe I'm missing a use case for always having a default excluded directory?

src/mobu/services/business/notebookrunner.py
```diff
@@ -87,14 +93,22 @@ async def clone_repo(self) -> None:
         with self.timings.start("clone_repo"):
             await self._git.clone("-b", branch, url, str(self._repo_dir))
 
+    def is_excluded(self, notebook: Path) -> bool:
+        # A notebook is excluded if any of its parent directories are excluded
+        parents = set(notebook.parents)
+        excluded_parents = parents & self._exclude_paths
+        return len(excluded_parents) != 0
```
rra, line 100:
```suggestion
        return bool(set(notebook.parents) & self._exclude_paths)
```

tests/business/notebookrunner_test.py
```diff
@@ -76,7 +76,7 @@ async def test_run(
             "business": {
                 "failure_count": 0,
                 "name": "NotebookRunner",
-                "notebook": "test-notebook.ipynb",
+                "notebook": ANY,
```
rra, line 79: I think we've now lost this test, in that nothing is now testing that the `notebook` data is set correctly to the executing notebook. It's some duplication, but it might be a bit cleaner to add a separate test for the recursive notebook running that uses a separate directory.

tests/business/notebookrunner_test.py
```diff
@@ -59,7 +59,7 @@ async def test_run(
                     "options": {
                         "spawn_settle_time": 0,
                         "execution_idle_time": 0,
-                        "max_executions": 1,
+                        "max_executions": 10,
```
rra, line 62: I think this would be a stronger test if `max_executions` exactly matched the number of notebooks that should be run, since then you can also test for off-by-one errors (make sure the first notebook isn't run twice, for instance).

tests/business/notebookrunner_test.py
```diff
@@ -93,9 +93,134 @@ async def test_run(
...
+
+    # Set up git client
+    git = Git(repo=repo_path)
+    await git.init("--initial-branch=main")
+    await git.config("user.email", "gituser@example.com")
+    await git.config("user.name", "Git User")
+    for path in repo_path.iterdir():
+        if not path.name.startswith("."):
+            await git.add(str(path))
+    await git.commit("-m", "Initial commit")
```
rra, line 147: Since there are now a couple copies of this code, it's probably time to move it into a helper function.

tests/business/notebookrunner_test.py
```diff
@@ -93,9 +93,134 @@ async def test_run(
...
+                "name": "test",
+                "count": 1,
+                "user_spec": {"username_prefix": "testuser"},
+                "scopes": ["exec:notebook"],
+                "business": {
+                    "type": "NotebookRunner",
+                    "options": {
+                        "spawn_settle_time": 0,
+                        "execution_idle_time": 0,
+                        "max_executions": 10,
```
rra, line 165: Same note here about `max_executions`.

rra (commented):
Overall, looks good, thank you! Just a couple of notes and suggestions.

src/mobu/constants.py
fajpunk: Agreed it seems magical.  There are a couple of issues:
1. One of the things we want is for folks to be able to manage these excluded notebooks without having to make phalanx PRs.  IIUC, all of the flock config is defined in phalanx, so folks would have to make phalanx PRs as they add and remove excluded dirs (or make one phalanx MR to get their own single exclude dir like in this magic default)
2. A magic default exclude dir gets us out of having to tightly coordinate this mobu release with notebook repo changes.  With this, we can [modify the notebook repos](https://github.com/lsst-sqre/system-test/pull/84) any time before the mobu release to any environment without having to coordinate adding the `exclude_dirs` flock option at the same time as we release to each environment.
    * To address this, we could have this behavior be opt-in by adding a `run_in_subdirs` option that defaults to False.  Then we can deploy the mobu change without affecting current behavior.  We could then change the default at some point in the future when all of the flocks have their excluded dirs specified appropriately.

One bigger change we could make to address both of these issues is to have this behavior specified in a config file or something in the notebook repos themselves.  Then:
* We could first add that config file to the repos before deploying the mobu changes and current behavior would be kept everywhere
* No phalanx changes would be needed to manage these exclude dirs

tests/business/notebookrunner_test.py
fajpunk: Yeah, my change makes that test weaker, but I also wanted to ensure that no notebooks from excluded dirs would be run.  I changed `max_execution` back to the exact number of notebooks, but I also moved the "cycle done" message from the beginning of the _next_ cycle, to the end of `execute_code`, and I'm asserting that's logged.

That gets us the best of both worlds, right?

src/mobu/constants.py
fajpunk: I ditched the magic dir.  Seems not worth the magic and confusion.  I add the subdirs in the currently existing flocks in the [phalanx PR](https://github.com/lsst-sqre/phalanx/pull/3346).

fajpunk: And I think I misunderstood the way the release process will work.  The mobu version and the flock config will get synced together for each environment; there won't be a state where mobu is updated but the flock config isn't.

changelog.d/20240520_113001_danfuchs_notebook_directories.md
```diff
@@ -0,0 +1,10 @@
+<!-- Delete the sections that don't apply -->
+
+### Backwards-incompatible changes
+
+- NotebookRunner business now runs all notebooks in a repo by default except in the `mobu_exclude` or its descendants
```
rra, line 5: You dropped the `mobu_exclude` part. We've also been putting periods at the end of change log entries (sort of an arbitrary style thing).

changelog.d/20240520_113001_danfuchs_notebook_directories.md
```diff
@@ -0,0 +1,10 @@
...
+- Add `exclude_dirs` option to NotebookRunner business to list additional directories in which notebooks will not be run
+
+### Other changes
+
+- Updated python dependencies
```
rra, line 10: We don't note this in the change log because we do it with every release, so it just adds noise. Instead, we document at the start of CHANGELOG.md that we do this with every release.

rra (approved):
This looks great. Just a couple of minor nits on the change log entry.

PR #360, opened 2024-07-19
DM-45079: Move test data

fajpunk:
Move all notebooks to the test data dir.

This is a subset of the [service-specific work](https://github.com/lsst-sqre/mobu/pull/359)


rra (approved):
Looks great, and thank you for the small PR!

PR #361, opened 2024-07-19
DM-45709: Don't run notebooks in environments that don't have the necessary services

fajpunk:
* Parse a 'mobu' section out of notebook metadata
* Add a list of available services to app config (to be replaced by some kind of service discovery at some point)
* Don't run the notebook if it declares required services in the metadata and the services aren't available in the environment.

Putting the required services in the notebook metadata can be done in JupyterLab, though is not the nicest UX for this. We now have an in-repo config file (mobu.yaml), maybe it would be better to keep a manifest in there? I like having it closer to the notebook though. I'm open to other suggestions.

### DIFF
--- a/changelog.d/20240719_150741_danfuchs.md
+++ b/changelog.d/20240719_150741_danfuchs.md
@@ -1,0 +1,5 @@
+<!-- Delete the sections that don't apply -->
+
+### New features
+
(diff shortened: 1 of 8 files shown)


rra (approved):
This change looks good! A couple of general notes:
- You can set `model_config = SettingsConfigDict(env_prefix="MOBU_", case_sensitive=False)` in the `Config` model and then if the environment variable name is the same as the field name except in all caps with `MOBU_` prepended, you can leave off the explicit `validation_alias`.
- We should at some point (not in this PR) figure out a better test strategy that can tie execution results to specific notebooks and make each new test a bit more succinct rather than having to copy and paste large blocks of status information to compare against.

PR #363, opened 2024-07-23
DM-45079: Execute only non-filtered notebooks  when requesting explicit notebooks

fajpunk:
GitHub CI checks ask mobu to run an explicit list of notebooks. Some of these notebooks may be excluded once mobu picks up the job though.

This makes sure that these jobs runs the filtered notebooks in a single session.

Mobu could end up not actually running any notebooks, if all of the notebooks in the explicit list are excluded.  In this case, mobu will conclude the check as a success. A future PR will make mobu conclude the check as neutral in this case.

### DIFF
--- a/src/mobu/services/business/notebookrunner.py
+++ b/src/mobu/services/business/notebookrunner.py
@@ -196,7 +196,7 @@ class NotebookRunner(NubladoBusiness):
                 if not_found:
                     msg = (
                         "Requested notebooks do not exist in"
-                        " {self._repo_dir}: {not_found}"
+                        f" {self._repo_dir}: {not_found}"
                     )
                     raise NotebookRepositoryError(msg, self.user.username)
                 filter_results.excluded_by_requested = (
--- a/src/mobu/services/business/notebookrunner.py
+++ b/src/mobu/services/business/notebookrunner.py
@@ -24,6 +24,7 @@ from ...constants import GITHUB_REPO_CONFIG_PATH
 from ...exceptions import NotebookRepositoryError, RepositoryConfigError
 from ...models.business.notebookrunner import (
     ListNotebookRunnerOptions,
+    NotebookFilterResults,
     NotebookMetadata,
     NotebookRunnerData,
     NotebookRunnerOptions,
@@ -90,8 +91,15 @@ class NotebookRunner(NubladoBusiness):
     async def cleanup(self) -> None:
         shutil.rmtree(str(self._repo_dir))
         self._repo_dir = None
+        self._notebook_filter_results = None
 
     async def initialize(self) -> None:
+        """Prepare to run the business.
+
+        * Check out the repository
+        * Parse the in-repo config
+        * Filter the notebooks
+        """
         if self._repo_dir is None:
             self._repo_dir = Path(TemporaryDirectory(delete=False).name)
             await self.clone_repo()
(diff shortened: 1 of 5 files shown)


src/mobu/services/business/notebookrunner.py
```diff
@@ -149,46 +167,68 @@ def missing_services(self, notebook: Path) -> bool:
...
+        return filter_results
 
     def next_notebook(self) -> Path:
+        if not self._notebooks:
+            self._notebooks = self.find_notebooks()
         if not self._notebook_paths:
-            self._notebook_paths = self.find_notebooks()
+            self._notebook_paths = random.sample(
+                list(self._notebooks.runnable), k=len(self._notebooks.runnable)
+            )
```
rra, line 231:
```suggestion
            self._notebook_paths = list(self._notebooks.runnable)
            random.shuffle(self._notebook_paths)
```
It was confusing to see sample used here with k equal to the length of the list. I think the above is what you intended?

src/mobu/services/business/notebookrunner.py
```diff
@@ -53,7 +55,7 @@ class NotebookRunner(NubladoBusiness):
 
     def __init__(
         self,
-        options: NotebookRunnerOptions,
+        options: NotebookRunnerOptions | ListNotebookRunnerOptions,
```
rra, line 58: We have previously had a one-to-one correspondence between options and instances of a business, and while this is fairly simple, I kind of want to keep that pattern for conceptual simplicity for mobu overall. Would it make sense to have a base Nublado business and then have subclasses that implement the algorithm for determining which notebooks to run? I think that would make it easier to understand the overall structure.

This could be follow-on work in a separate PR, since it's going to be a bit of a rework.

src/mobu/services/business/notebookrunner.py
```diff
@@ -149,46 +167,68 @@ def missing_services(self, notebook: Path) -> bool:
             return True
         return False
 
-    def find_notebooks(self) -> list[Path]:
+    def find_notebooks(self) -> NotebookFilterResults:
```
rra, line 170: For future work, this method has gotten complicated enough that I would consider breaking it out into a separate service class. It feels like there's a simple API hiding in here and a reasonable amount of code that could be pushed into a different class and hidden.

rra (approved):
Overall, looks good to me!

src/mobu/services/business/notebookrunner.py
fajpunk: The only reason I split these out was because it only makes sense to specify one of `notebooks_to_run` and `max_executions` at the same time.
What about just going back to a single Options, and using a model validator to validate that only one of those options is specified?

rra: I dunno, I think having them separate is good! I have separated config and business for similarly trivial things before (like the two TAP runners). I think the two child classes may have only one or two very short methods. But maybe I'm more comfortable with large inheritance hierarchies than I should be.

I don't have any strong feelings about this, to be clear; if you think it's better the way it is, I'm fine with leaving it alone. It just jumped out at me since it felt inconsistent with the pattern for other businesses.

src/mobu/services/business/notebookrunner.py
fajpunk: Agreed, captured in the [cleanup ticket](https://rubinobs.atlassian.net/browse/DM-45394)

src/mobu/services/business/notebookrunner.py
fajpunk: The base-business-with-subclasses pattern does seem to fit better.  I don't have super strong feelings either, glad to refactor it later, but yeah, I'll leave it as-is now.

PR #374, opened 2024-10-17
tickets/DM-46480: adopt shared rubin-nublado-client

athornton:
Rather than use internal client, use the shared client from the `nublado` repo.

This is currently dependent on DM-46480 branch of nublado, and will be changed once that work is merged and released.


src/mobu/exceptions.py
```diff
@@ -542,6 +885,72 @@ def to_slack(self) -> SlackMessage:
...
+        -------
+        JupyterWebSocketError
+            Converted exception.
+        """
+        anno: dict[str, str] = {}
+        if annotations is not None:
+            anno = annotations
+        body = exc.body
+        if body is not None:
+            body_bytes = body.encode()
```
fajpunk, line 936: Is it worth storing the encoded body on the nublado `JupyterWebSocketException`?

src/mobu/services/business/nublado.py
```diff
@@ -166,6 +185,27 @@ async def startup(self) -> None:
...
+        try:
+            await self._execute()
+        except Exception as exc:
+            monkey = None
+            event = "execute_code"  # Fallback
+            if isinstance(exc, MobuMixin):
+                if exc.monkey:
+                    monkey = exc.monkey
+                if exc.event:
+                    event = exc.event
```
fajpunk, line 197: What about something like:
```suggestion
            monkey = getattr(exc, "monkey", None)
            event = getattr(exc, "event", "execute_code")
```

src/mobu/services/business/nublado.py
```diff
@@ -202,13 +242,36 @@ async def idle(self) -> None:
...
+        with self.timings.start("hub_login", self.annotations()) as sw:
+            try:
+                await self._client.auth_to_hub()
+            except ClientJupyterProtocolError as exc:
+                raise JupyterProtocolError.from_client_exception(
+                    exc,
+                    event=sw.event,
+                    annotations=sw.annotations,
+                    started_at=sw.start_time,
+                ) from exc
```
fajpunk, line 254: Is this adding new info, the event, annotations, and start time, to what was in this exception before this change?  Or was that somehow getting in there before?

tests/autostart_test.py
```diff
@@ -136,10 +136,8 @@ async def test_autostart(client: AsyncClient, jupyter: MockJupyter) -> None:
                 "business": {
                     "failure_count": 0,
                     "image": {
-                        "description": "Recommended (Weekly 2077_43)",
-                        "reference": (
-                            "lighthouse.ceres/library/sketchbook:recommended"
-                        ),
+                        "description": ANY,
+                        "reference": ANY,
```
fajpunk, line 140: Why do we lose the ability to test specific descriptions and references here?

fajpunk: (and elsewhere in these tests)

fajpunk (commented):
My main architectural questions are:
* Would it be cleaner to keep a nublado support in mobu, whose only job is to wrap the extracted nublado client's methods and do the exception transormations, instead of doing those at each call site?  Would that even be possible?
* Is there any way to extract out the common stuff we're doing in all of the wrapped exceptions, like adding annotations and mobu fields, somehow so we don't have to repeat the same logic in every wrapped exception?

src/mobu/exceptions.py
```diff
@@ -351,36 +376,70 @@ def __str__(self) -> str:
         else:
             msg = f"{self.user}: running {self.code_type} failed"
         if self.error:
-            msg += f"\nError: {_remove_ansi_escapes(self.error)}"
+            if self.error:
```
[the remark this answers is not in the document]
athornton: no, we do not

src/mobu/exceptions.py
```diff
@@ -186,14 +229,34 @@ def __init__(
...
+    @classmethod
+    def from_slack_exception(cls, exc: NubladoClientSlackException) -> Self:
+        return cls(
+            msg=exc.message,
+            user=exc.user,
+            started_at=exc.started_at,
+            failed_at=exc.failed_at,
+        )
+
+    def common_fields(self) -> list[SlackBaseField]:
```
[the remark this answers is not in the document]
athornton: Oh nice I didn't know about @override

src/mobu/exceptions.py
```diff
@@ -351,36 +376,70 @@ def __str__(self) -> str:
...
-            blocks=self.common_blocks(),
-            attachments=attachments,
+    @classmethod
+    def from_client_exception(
+        cls,
+        exc: ClientCodeExecutionError,
+        monkey: str | None = None,
+        event: str | None = None,
+        annotations: dict[str, str] | None = None,
+        started_at: datetime.datetime | None = None,
```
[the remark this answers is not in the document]
athornton: `failed_at` gets tacked in as `now` if it's missing way down in the Safir level, IIRC.

src/mobu/exceptions.py
```diff
@@ -389,15 +448,168 @@ class GitHubFileNotFoundError(Exception):
...
+        self.mobu_init(monkey=monkey, event=event)
+
+    @classmethod
+    def from_client_exception(
+        cls,
+        exc: ClientJupyterProtocolError,
+        event: str | None = None,
+        monkey: str | None = None,
+        started_at: datetime.datetime | None = None,
+        annotations: dict[str, str] | None = None,
```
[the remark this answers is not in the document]
athornton: Currently, annotations are never set at construction time of the bottom base clase--they are initialized to the emtpy dict.

src/mobu/services/business/nublado.py
athornton: Well, the problem is that there was no Nublado Client distinct from mobu, and hence no client exceptions, and factoring the two layers apart got messy.

tests/autostart_test.py
athornton: Because it wasn't clear to me that we would want to set them directly or use the ones from the NubladoClient testing suite, and there they are not exposed well.  I think.

athornton:
The right way to address the architectural concerns (and this is something I talked about a little with @rra ) is ultimately to turn almost all of these exception fields into annotations.  That will weaken the typing a little, but will make composing exceptions a whole lot easier.

It's a little subtle, in that we want to maintain the one-column vs two-column distinctions (that is, blocks versus fields) and large amounts of data will need to be packaged as attachments, and that (or something near there) is not documented in the Slack API...and then all of that really needs to be in Safir, neither in Mobu nor in NubladoClient.

But that's going to be a lot of work, so what I want to do in the immediate term is to land splitting out the NubladoClient into the Nublado monorepo, and adopting it into ghostwriter, mobu, and noteburst.

In addition to being a lot of work, the annotation-based Safir rewrite is work we should not undertake without Jonathan, who will have strong opinions about how Safir ought to work, so we really can't tackle it until 2025 when he's back.

sqreadmin:
I guess it's the moral equivalent of _convert_exception() in the Nublado Client.  Hm.

sqreadmin:
Waaaaait why am I logged in as sqreadmin?

src/mobu/exceptions.py
athornton: Imma say "not at this point", although when we do the big Slack Exception refactor, the answer there is "yes" or maybe even better, "we'll get it for free because that's how the restructure will work".

src/mobu/exceptions.py
```diff
@@ -69,7 +88,31 @@ def _remove_ansi_escapes(string: str) -> str:
...
 
 
-class GafaelfawrParseError(SlackException):
+class MobuMixin:
+    """Mixin class to add `event` and `monkey` fields to Exception."""
+
+    def __init__(
+        self, event: str | None = None, monkey: str | None = None
+    ) -> None:
+        self.mobu_init()
```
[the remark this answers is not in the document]
athornton: I think so.  If you want to set the monkey and event at initialization....

....Although...you wouldn't ever use it standalone, that's the point of a Mixin, innit?

src/mobu/exceptions.py
```diff
@@ -389,15 +448,168 @@ class GitHubFileNotFoundError(Exception):
...
+        else:
+            message = "Spawning lab failed"
+        super().__init__(
+            message, user, started_at=started_at, failed_at=failed_at
+        )
+        self.log = log
+        self.mobu_init(monkey=monkey, event=event)
+
+    @classmethod
+    def from_client_exception(
```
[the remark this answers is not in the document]
athornton: IDK.  In most cases, from_exception() is how the base class turns *its* parent exception into the NubladoClientException.  We'd be changing the call signature if we made it an override.

My tendency is to pick a different name, but maybe I'm wrong about that?

src/mobu/services/business/nublado.py
athornton: I *think* that for all of these exceptions, we will likely already be supplying a start time, but since we have it as part of the stopwatch there's no harm in specifying it.

src/mobu/exceptions.py
```diff
@@ -491,28 +485,28 @@ def from_client_exception(
...
         """
-        anno: dict[str, str] = {}
-        if annotations is not None:
-            anno = annotations
         new_exc = cls(
             msg=exc.message,
             user=exc.user,
             started_at=started_at or exc.started_at,
-            failed_at=exc.failed_at,
+            failed_at=exc.failed_at or exc.failed_at,
```
fajpunk, line 501:
```suggestion
            failed_at=failed_at or exc.failed_at,
```

fajpunk, line 501:
```suggestion
            failed_at=failed_at or exc.failed_at,
```

fajpunk (commented):
One more typo, but yeah, I guess we can put off a refactor until we do the bigger refactor

fajpunk (approved):

PR #385, opened 2024-11-26
DM-47811: Fixes to address FAST001 and FAST002 ruff rules and update-deps #232

stvoutsin:
Fix ruff rules [FAST001](https://docs.astral.sh/ruff/rules/fast-api-redundant-response-model/) & [FAST002](https://docs.astral.sh/ruff/rules/fast-api-non-annotated-dependency/) which were introduced in ruff version [0.8.0](https://github.com/astral-sh/ruff/releases/tag/0.8.0)

This PR addresses this in the following way:

- Updates the dependencies including ruff
- Removes the response_model param which is no longer needed
- A few changes to the order of imports which are autofixes


changelog.d/20241126_224636_steliosvoutsinas.md
rra, line 1: No change log necessary for these since there are no user-visible changes.

rra (approved):
Looks good minus the change log. Thanks!

changelog.d/20241126_224636_steliosvoutsinas.md
stvoutsin: Ah ok i see was using it as documentation for any changes. I've removed the entry

PR #387, opened 2024-12-10
DM-47389: app metrics

fajpunk:
Publish app metrics events.

[Phalanx support](https://github.com/lsst-sqre/phalanx/pull/3989)


src/mobu/services/business/gitlfs.py
```diff
@@ -65,10 +75,23 @@ async def execute(self) -> None:
...
+            elapsed = sw.elapsed.total_seconds()
+            self.logger.info(f"...Git-LFS check finished after {elapsed}s")
+
+            event.duration_total = sw.elapsed
+            event.success = True
+        except:
+            event.success = False
+            raise
+        finally:
+            await ed.events.git_lfs_check.publish(event)
```
fajpunk, line 94: Is this directly-accessing-the-events-dependency-singleton pattern horrible? vs. threading it through the process context and injecting it as a parameter into the business constructor?

src/mobu/services/business/gitlfs.py
```diff
@@ -65,10 +75,23 @@ async def execute(self) -> None:
         created anew.
         """
         self.logger.info("Running Git-LFS check...")
-        with self.timings.start("execute git-lfs check") as sw:
-            await self._git_lfs_check()
-        elapsed = sw.elapsed.total_seconds()
-        self.logger.info(f"...Git-LFS check finished after {elapsed}s")
+        event = GitLfsCheck(
+            success=False,
+            **self.common_event_attrs(),
```
rra, line 80: Nit: If you remove the comma, this all fits on one line.

src/mobu/services/business/gitlfs.py
rra: Yeah, there are a few things that bother me about this: accessing the dependency directly, the renaming of imports (I don't like doing that unless one has to because multiple modules have symbols with the same name), and using a really short name that is kind of meaningless if one is looking at just a portion of the file. I tend to stick pretty closely to dependency injection so that it's part of the business as `self.events`, which is almost as short and I think somewhat clearer.

src/mobu/services/business/notebookrunner.py
```diff
@@ -295,26 +315,88 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
-                    cell_number=f"#{cell['_index']}",
-                    cell_source=code,
-                )
-                await self.execute_cell(session, code, cell_id, ctx)
-                if not await self.execution_idle():
-                    break
+            with self.timings.start(
+                "execute_notebook", self.annotations(self._notebook.name)
+            ) as sw:
+                try:
```
rra, line 321: This is not actionable feedback, just me musing, but something feels weird here. It feels like we're running timings and events in parallel, and are doing a lot of manual bookkeeping that creates deeply nested structure. My brain wants to find a new abstraction that unifies these layers. Something like... starting an event that can succeed or fail with a context manager, having it keep a top-level duration, and then possibly having it expose additional context managers that time specific subtasks, and then have the `__exit__` dunder method of the context manager capture success or failure based on whether an exception was thrown?

Happy to merge this without that change, but I feel like we're getting a bit of complexity creep here. Maybe a bunch of it goes away when we throw out the old way of recording timings?

src/mobu/services/business/notebookrunner.py
```diff
@@ -295,26 +315,88 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
+                duration=duration,
+                success=True,
+            )
+        )
+
+    async def _publish_notebook_failure(self) -> None:
+        await ed.events.notebook_execution.publish(
+            NotebookExecution(
+                **self.common_notebook_event_attrs(),
+                duration=None,
```
rra, line 359: We definitely want to record duration for failure as well. That's been the only way that I've been able to figure out that the failure was due to a timeout.

src/mobu/services/business/notebookrunner.py
```diff
@@ -295,26 +315,88 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
+            )
+        )
+
+    async def _publish_cell_failure(
+        self, *, cell_id: str, contents: str
+    ) -> None:
+        await ed.events.notebook_cell_execution.publish(
+            NotebookCellExecution(
+                **self.common_notebook_event_attrs(),
+                duration=None,
```
rra, line 383: Likewise here, we want to record duration on failure.

src/mobu/services/business/nubladopythonloop.py
```diff
@@ -30,22 +34,57 @@ class NubladoPythonLoop(NubladoBusiness):
...
+                code=code,
+                success=True,
+                **self.common_event_attrs(),
+            )
+        )
+
+    async def _publish_failure(self, code: str) -> None:
+        await ed.events.nublado_python_execution.publish(
+            NubladoPythonExecution(
+                duration=None,
```
rra, line 85: And likewise here on duration for failure.

tests/business/gitlfs_test.py
```diff
@@ -49,6 +53,32 @@ async def test_run(
     assert "Running Git-LFS check..." in r.text
     assert "Git-LFS check finished after " in r.text
 
+    published = cast(MockEventPublisher, ed.events.git_lfs_check).published
```
rra, line 56: I have a mild preference for something like:
```suggestion
    assert isinstance(ed.events.git_lfs_check, MockEventPublisher)
    published = ed.events.git_lfs_check.published
```
although I don't feel strongly about it.

rra (commented):
I'm not seeing metrics for the lab spawning process, just for notebook execution. Can we get those as well? I think those are going to be really important. (I don't care about metrics for login and certainly not about the intentional pauses, or for things like creating a web socket.)

tests/business/gitlfs_test.py
rra: Also, I wonder if there's any way that we can get rid of having to do either the assertions or the casts. I feel like there's something here in the type system, but it's not at all obvious how to go about it given how the publishers are structured right now. I'll ponder.

fajpunk:
> I'm not seeing metrics for the lab spawning process, just for notebook execution. Can we get those as well? I think those are going to be really important. (I don't care about metrics for login and certainly not about the intentional pauses, or for things like creating a web socket.)

Yeah, good call, this is probably one of the most useful events we can have! I added `NubladoSpawnLab` and `NubladoDeleteLab` events.

src/mobu/events.py
```diff
@@ -0,0 +1,100 @@
...
+    success: bool
+
+
+class NotebookCellExecution(NotebookBase):
+    """Reported after a notebook cell is finished executing."""
+
+    duration: timedelta | None
+    cell_id: str
+    success: bool
+    contents: str
```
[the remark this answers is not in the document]
fajpunk: Agreed that a commit hash would be more useful than storing the actual contents. If we wanted to query by that hash, either in a where clause or in a group by, it would have to be an Influx tag, and that could contribute to high cardinality, but it's worth a shot.

I don't think a hash is useful if we're not going to be able to make it a tag. I imagine that the cell execution timings would mainly be useful if we see a big change in the notebook execution timings, and then we would want to see durations/errors by hash on both the notebook and cell execution events.

docs/development/github.rst
```diff
@@ -36,7 +36,7 @@ You can use https://smee.io to proxy GitHub webhook requests from the data-dev M
        pyyaml
       -safir>=5.0.0
       +# safir>=5.0.0
-      +-e /home/danfuchs/src/safir
+      +-e /home/danfuchs/src/safir/safir
```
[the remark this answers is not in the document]
fajpunk: Updated to describe that option, though in this particular case, I don't think we'd ever want to actually release this Safir change.

src/mobu/events.py
```diff
@@ -0,0 +1,100 @@
...
+    duration_git_attribute_installation: timedelta | None = None
+
+
+class TapQuery(EventBase):
+    """Reported when a TAP query is executed."""
+
+    success: bool
+    duration: timedelta | None
+    sync: bool
+    query: str
```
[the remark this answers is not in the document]
fajpunk: I think your intuition is correct. Folks can query by flock and user and then look in the phalanx mobu config to find out what the queries currently are, but if they're investigating a significant change in duration, they'll have to go through the git history manually.

tests/business/gitlfs_test.py
fajpunk: One way was [having a property on the interface fails at runtime if it's not available](https://github.com/lsst-sqre/safir/pull/348#discussion_r1874174707)

Another option I considered with the test helpers was a pytest fixture that mokeypatches the `publish` method of `EventPublisher` and yields a `PublishedList`. I didn't choose that for a couple of reasons:
* I was scared that the patching would be brittle and could break depending on how folks organized their apps, but this is probably a unfounded fear
* I didn't want to patch over too much actual functionality. Right now the only other thing that a publisher does is construct the event, I wanted as much running in the tests as possible.
* I thought it might be a pain to keep the patching and mock logic in sync with the events code as it changed.  We have to keep the `MockEventPublisher` in sync too, but it's right there in your face if you're changing anything, vs having the patch and mock code live somewhere else entirely.

But I could probably be convinced that the mock and patch is a better way to do this.

fajpunk: > I have a mild preference for something like: assert isinstance(ed.events.git_lfs_check, MockEventPublisher)

`isinstance` won't work if you want to narrow the `PublishedList` to a specific event type:

```python
   pub_cast_narrowed = cast(
        MockEventPublisher[NotebookExecution], ed.events.notebook_execution
    ).published
    pub_cast_narrowed.assert_published_all([common])
    event = pub_cast_narrowed[0]
    assert event.success  # Good
    assert event.asdfasdfasdf  # Bad
    # yay! mypy: error: "NotebookExecution" has no attribute "asdfasdfasdf"
```

```python
    assert isinstance(
        ed.events.notebook_execution, MockEventPublisher[NotebookExecution]
    )
    # mypy: error: Parameterized generics cannot be used with class or instance checks  [misc]
```

In all of these cases, though, I guess  it doesn't matter because I'm never accessing the EventPayloads directly.

fajpunk: If you're only mildly offended, I'll leave the `cast` as-is to align with [the safir docs](https://safir.lsst.io/user-guide/metrics/index.html#unit-testing), though the safir docs could be updated with a narrowed cast example.

src/mobu/services/business/gitlfs.py
fajpunk: [Here](https://github.com/lsst-sqre/mobu/pull/387/commits/bc86c8c379aade7608509d8aadef16aac652efa8) is this change to access it through the `ProcessContext`, I think that's the right place to put it in Mobu

tests/business/gitlfs_test.py
rra: Note that cast doesn't really do that either in the sense that it doesn't check, it just tells mypy to blindly assume that this is the case. I usually prefer `isinstance` because it checks, although it's a good point that this doesn't check the subtype (which I don't see a good way to check either way).

Anyway, nah, I don't feel strongly about it. It just all feels kind of awkward and I'm going to keep pondering whether there's some better approach. I do see why originally you added the abstract property, but I feel like keeping the production API clean is worth a bit of typing lossage in tests.

src/mobu/events.py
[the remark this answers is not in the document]
rra: I think I'm missing something: can't we dump the timings and the hashes by notebook and cell and see exactly when the timing degraded? In practice it's probably going to be rather visible on a graph against time, and then we just look at the data point before and after the change and see what hashes they used and that gives us the commit. I can see how this would be a bit easier if it were a tag, but it doesn't seem that important to make it a tag.

fajpunk:
I'll bring up in co-work whether we want to replace the current timing stuff with sampled Sentry tracing. If we do decide to do that, we may not want some of these as analytical metrics.  Notebook cell executions for example, I don't see the value in keeping that data forever, but I do see the value in keeping it short-term to debug why an entire notebook is suddenly slow

tests/business/gitlfs_test.py
fajpunk: > Note that cast doesn't really do that either in the sense that it doesn't check, it just tells mypy to blindly assume that this is the case.

Definitely, though the test would then blow up at runtime if you told mypy wrong. Agreed that this is better in the tests than in the app. I'll keep thinking about it too!

rra (approved):
This looks good, thanks! I'm fairly sure that there are ways to make some of this code simpler, or at least with somewhat sharper separation of concerns, but I think that's an orthogonal refactoring to the work you're doing, and as you say we should probably wait until we've integrated Sentry and then see what the remaining complexity looks like.

PR #391, opened 2025-01-06
DM-48101: Sentry for notifications and timings

fajpunk:
* Replace Safir `SlackException` functionality with Sentry error reporting.
* Replace `timings` functionality with Sentry tracing.

The biggest conceptual change is that instead of putting error details on the exception objects themselves, they get put into the per-business-execution Sentry scope, and whatever is in the scope is sent to Sentry when an error occurs (except for a very few cases where we still attach info directly to exceptions).

Instruments custom scopes and transactions and fingerprints, which are described in the new `sentry.rst` file in the docs.

New config options are updated in Phalanx in https://github.com/lsst-sqre/phalanx/pull/4092

Finally, this uses some more general Sentry helpers proposed for Safir in https://github.com/lsst-sqre/safir/pull/366.


src/mobu/sentry.py
```diff
@@ -0,0 +1,114 @@
+"""Helpers for sentry instrumentation."""
+
+from __future__ import annotations
+
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any, Literal
+
+import sentry_sdk
+from astropy.utils.xml.writer import contextlib
```
jonathansick, line 10: Is this an autocomplete typo?

fajpunk: Yes :(

src/mobu/exceptions.py
```diff
@@ -169,927 +122,91 @@ def __init__(self, monkey: str) -> None:
...
-        field = SlackTextField(heading="Expected", text=self.expected)
-        message.fields.append(field)
-        field = SlackTextField(heading="Received", text=self.received)
-        message.fields.append(field)
-        return message
+class JupyterSpawnTimeoutError(Exception):
+    """Timed out waiting for the lab to spawn."""
+
+
+class JupyterDeleteTimeoutError(Exception): ...
```
rra, line 205: Needs a doc string.

src/mobu/exceptions.py
```diff
@@ -169,927 +122,91 @@ def __init__(self, monkey: str) -> None:
...
+
+
+class JupyterSpawnError(Exception):
+    """The Jupyter Lab pod failed to spawn."""
+
+
+class NotebookCellExecutionError(Exception): ...
```
rra, line 212: Needs a doc string.

src/mobu/main.py
```diff
@@ -38,6 +41,14 @@
 __all__ = ["create_app", "lifespan"]
 
 
+if os.environ.get("MOBU_SENTRY_DSN"):
+    sentry_init(
+        dsn=config_dependency.config.sentry_dsn,
+        env=config_dependency.config.sentry_environment,
+        traces_sample_config=config_dependency.config.sentry_traces_sample_config,
+    )
```
rra, line 49: This is mostly a style thing, but I'd put this inside `create_app` so that we're not running some code on module import and other code on uvicorn startup unless there's some reason why they really need to happen at different times.

src/mobu/services/business/base.py
```diff
@@ -199,7 +196,7 @@ async def run_once(self) -> None:
     async def idle(self) -> None:
         """Pause at the end of each business loop."""
         self.logger.info("Idling...")
-        with self.timings.start("idle"):
+        with start_span(op="idle"):
             await self.pause(self.options.idle_time)
```
rra, line 200: Should we drop all the spawns around idling? The only reason why those were being captured before was that we wanted to see if counting up all the timers roughly equaled the elapsed time to make sure that we weren't avoiding timing any major operations.

src/mobu/services/business/notebookrunner.py
```diff
@@ -316,42 +324,63 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
+            # don't, the nested transaction shows up as "No instrumentation" in
+            # the enclosing transaction in the Sentry UI.
+            with start_span(
+                name="execute_notebook", op="execute_notebook"
+            ) as notebook_span:
+                notebook_span.set_data("notebook_info", notebook_info)
+                with start_transaction(
+                    name="execute_notebook",
+                    op="execute_notebook",
+                ) as span:
```
rra, line 351: This code has wandered into a really high indentation level. I think it would be worth refactoring this into helper methods for clarity.

src/mobu/services/business/notebookrunner.py
```diff
@@ -392,28 +421,49 @@ async def execute_cell(
...
+                        )
                 await self._publish_cell_event(
                     cell_id=cell_id,
-                    duration=sw.elapsed,
+                    duration=duration(span),
                     success=False,
                 )
-                raise
+                raise NotebookCellExecutionError(
+                    f"{getattr(context, 'notebook', '<unknown notebook')}:"
```
rra, line 460: This is pretty complex code to put in an f-string. I'd lift it into a variable.

src/mobu/services/business/nublado.py
```diff
@@ -157,57 +158,36 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
-            except JupyterTimeoutError:
-                msg = "Unable to delete pre-existing lab, continuing anyway"
-                self.logger.warning(msg)
+        # We need to start a span around this transaction becaues if we don't,
+        # the nested transaction shows up as "No instrumentation" in the
+        # enclosing transaction in the Sentry UI.
+        with sentry_sdk.start_span(name="startup", op="startup"):
+            with start_transaction(
+                name=f"{self.__class__.__name__} - startup",
+                op=f"{self.__class__.__name__} - startup",
```
rra, line 168: I'm confused about why the class name is included here when it's not included in the span name. Is the display of the transactions more ambiguous somehow?

src/mobu/services/business/nublado.py
```diff
@@ -157,57 +158,36 @@ async def execute_code(self, session: JupyterLabSession) -> None:
...
+                await self.hub_login()
+                if not await self._client.is_lab_stopped():
+                    try:
+                        await self.delete_lab()
+                    except JupyterDeleteTimeoutError:
+                        msg = (
+                            "Unable to delete pre-existing lab, continuing"
+                            " anyway"
+                        )
+                        self.logger.warning(msg)
```
rra, line 185: Code has gotten really indented here as well, enough so that it's starting to distort error messages. I usually take that as a sign to refactor into helper methods.

src/mobu/services/business/nublado.py
```diff
@@ -224,76 +204,60 @@ async def execution_idle(self) -> bool:
...
     async def shutdown(self) -> None:
-        await self.hub_login()
-        await self.delete_lab()
+        # We need to start a span around this transaction becaues if we don't,
+        # the nested transaction shows up as "No instrumentation" in the
+        # enclosing transaction in the Sentry UI.
+        with sentry_sdk.start_span(name="shutdown", op="shutdown"):
+            with start_transaction(
+                name=f"{self.__class__.__name__} - shutdown",
+                op=f"{self.__class__.__name__} - shutdown",
```
rra, line 217: Same question here about the class name.

src/mobu/services/business/nublado.py
```diff
@@ -310,89 +274,55 @@ async def _spawn_lab(self, sw: Stopwatch) -> bool:  # noqa: C901
...
         self.logger.info("Creating lab session")
         opts = {"max_websocket_size": self.options.max_websocket_message_size}
-        stopwatch = self.timings.start("create_session", self.annotations())
+        create_session_cm = start_span(op="create_session")
+        create_session_cm.__enter__()
         async with self._client.open_lab_session(notebook, **opts) as session:
-            stopwatch.stop()
-            with self.timings.start("execute_setup", self.annotations()):
+            create_session_cm.__exit__(None, None, None)
+            with start_span(op="execute_setup"):
```
rra, line 316: This is pretty ugly. I think it would be better to wrap the Nublado client instead (or, maybe better, add Sentry support directly to the Nublado client?) so that the span management happens inside the async context manager returned by `open_lab_session`.

src/mobu/services/business/nublado.py
```diff
@@ -409,8 +339,15 @@ async def setup_session(self, session: JupyterLabSession) -> None:
             reference=reference.strip() if reference else None,
             description=description.strip() if description else None,
         )
+        set_tag(
+            "image",
+            self._image.description
+            or self._image.reference
+            or "<image unknown>",
+        )
```
rra, line 347: These are separate things. I would instead set separate tags for the image description and the image reference.

src/mobu/services/business/nubladopythonloop.py
```diff
@@ -58,15 +64,23 @@ def __init__(
...
+        sentry_sdk.start_span()
         for _count in range(self.options.max_executions):
-            with self.timings.start("execute_code", self.annotations()) as sw:
+            with start_span(op="execute_code") as span:
                 try:
                     reply = await session.run_python(code)
-                except:
+                except Exception as e:
+                    if isinstance(e, CodeExecutionError):
+                        if e.error:
```
rra, line 75:
```suggestion
                    if isinstance(e, CodeExecutionError) and e.error:
```

src/mobu/exceptions.py
```diff
@@ -169,927 +122,91 @@ def __init__(self, monkey: str) -> None:
...
-class TAPClientError(MobuSlackException):
-    """Creating a TAP client failed."""
-
-    def __init__(self, exc: Exception, *, user: str) -> None:
-        if str(exc):
-            error = f"{type(exc).__name__}: {exc!s}"
-        else:
-            error = type(exc).__name__
-        msg = f"Unable to create TAP client: {error}"
-        super().__init__(msg, user)
```
rra, line 993: You still need this; it's doing a bunch of work to add context to the exception.

src/mobu/services/business/tap.py
```diff
@@ -173,15 +170,12 @@ def _make_client(self, token: str) -> pyvo.dal.TAPService:
...
-            raise TAPClientError(e, user=self.user.username) from e
+        s = requests.Session()
+        s.headers["Authorization"] = "Bearer " + token
+        auth = pyvo.auth.AuthSession()
+        auth.credentials.set("lsst-token", s)
+        auth.add_security_method_for_url(tap_url, "lsst-token")
+        auth.add_security_method_for_url(tap_url + "/sync", "lsst-token")
+        auth.add_security_method_for_url(tap_url + "/async", "lsst-token")
+        auth.add_security_method_for_url(tap_url + "/tables", "lsst-token")
+        return pyvo.dal.TAPService(tap_url, auth)
```
rra, line 181: That try/catch block is there from hard-won experience, I fear. This change will cause it to throw random things like `ValueError` and `KeyError` with absolutely no useful information about what precisely failed. It's there to get the context of "this happened while creating a TAP client" into the exception along with the underlying exception type and message.

src/mobu/services/monkey.py
```diff
@@ -216,15 +187,22 @@ async def run_once(self) -> str | None:
...
-            await self.business.run_once()
-            self._state = MonkeyState.FINISHED
-        except Exception as e:
-            msg = "Exception thrown while doing monkey business"
-            self._logger.exception(msg)
-            error = str(e)
-            self._state = MonkeyState.ERROR
-        return error
+        with sentry_sdk.isolation_scope():
+            sentry_sdk.set_user({"username": self._user.username})
```
rra, line 191: If we have this, do we need the `gafaelfawr_user` thing that I asked about earlier?

src/mobu/services/monkey.py
```diff
@@ -216,15 +187,22 @@ async def run_once(self) -> str | None:
...
+            sentry_sdk.set_tag("business", self.business.__class__.__name__)
+            try:
+                with start_transaction(
+                    name=self.__class__.__name__,
+                    op=self.__class__.__name__,
+                ):
```
rra, line 197: I think these were supposed to be `self.business.__class__.__name__`. Right now, isn't this always the literal string `Monkey`? (If I'm right, it would also be worth refactoring that value into a variable; it would make the code easier to read.

rra (commented):
This is loads of work -- thank you so much!

I admit I'm a bit nervous about how Sentry-specific this all is, but I think there's no real way to avoid that and I guess it's just a mass search and replace to switch to any other reporting framework with a similar API. But I'm a bit sad that the context isn't carried by the exceptions themselves, since that would guarantee that anything Python-based would be able to use that data. This relies a lot more on global state, which just feels a touch more awkward. I suppose it does save laboriously adding all of the context to every exception we throw, though, so there is that.

src/mobu/exceptions.py
```diff
@@ -2,49 +2,31 @@
...
-    "MobuSlackException",
-    "MobuSlackWebException",
     "MonkeyNotFoundError",
     "SubprocessError",
-    "TAPClientError",
 ]
 
 
-def _remove_ansi_escapes(string: str) -> str:
+def remove_ansi_escapes(string: str) -> str:
```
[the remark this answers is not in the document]
rra: Or `NubladoBusiness`, rather, having looked at the whole diff.

src/mobu/services/business/nublado.py
rra: Having read the rest of the diff, I think I'm less confused. (I tend to write that as `type(class).__name__`, which is a bit more succinct.)

src/mobu/main.py
fajpunk: That should be fine, the only reason we'd want it at module level is if we wanted to catch exceptions in some other module-level logic happening in the module, of which there is none here.

src/mobu/services/business/base.py
fajpunk: I think it's nice to see the idle spans in the trace, instead of the time being included into whatever the containing span is.

src/mobu/services/business/tap.py
fajpunk: There's a `phase: make_client` tag set here (and a span created), but it's not easy to see because it's in the `startup` method instead.  I moved it into this method, does that make this better? You could see this in the issue in the Sentry UI. We could also put the `phase` tag in the sentry notification (though it's still not as clear as putting it in the exception message itself).

src/mobu/exceptions.py
fajpunk: We should have all of this context already in the Sentry context:
* started_at
   This is now in the `phase` Sentry context (via the `capturing_sentry_span` context manager)
* failed_at
   This is now the time of the Sentry event
* monkey
   This is a Sentry tag
* event
   This is now in the `phase` Sentry context (via the `capturing_sentry_span` context manager)

https://github.com/lsst-sqre/mobu/blob/main/src/mobu/exceptions.py#L205

src/mobu/services/business/nublado.py
fajpunk: Yeah, I agree that this should probably be in the nublado client itself, and I can add a ticket to do that. This is just a "don't make it worse" direct conversion for now :(

src/mobu/exceptions.py
```diff
@@ -123,25 +86,15 @@ def from_exception(
...
     ) -> None:
-        super().__init__(message, user)
-        self.error = error
-
-    @override
-    def to_slack(self) -> SlackMessage:
-        """Convert to a Slack message for Slack alerting.
+        super().__init__(message)
+        if user:
+            self.tags["gafaelfawr_user"] = user
```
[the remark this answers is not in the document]
fajpunk: Removed as unneeded per your [other comment](https://github.com/lsst-sqre/mobu/pull/391#discussion_r1926230398)

docs/development/sentry.rst
```diff
@@ -0,0 +1,34 @@
...
+`Isolation scopes <https://docs.sentry.io/platforms/python/enriching-events/scopes/>`_ and `transactions <https://docs.sentry.io/platforms/python/tracing/instrumentation/custom-instrumentation/#add-a-transaction>`_ are created manually.
+All tags and contexts are set on the isolation scope, and any exceptions are manually captured in the monkey runner loop.
+
+* There is one isolation scope for every execution of a business's ``run`` method.
+* There is one transaction for every execution of a business's ``run`` method.
+* For ``Nublado`` businesses, there are transactions created for startup and shutdown.
+* For ``NotebookRunner`` businesses, a transaction is created for every notebook execution.
+
+The details of the notebook execution transactions, or the startup and shutdown transactions, will not appear in the main business iteration transaction.
+We have these separate inner transactions because Sentry waits until all spans are closed in a transaction before sending it, and we don't want to wait for ``max_executions`` notebook executions before tracing information for a single notebook executions is sent.
```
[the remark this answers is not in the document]
fajpunk: Yep, having a transaction for the `run` method is definitely incorrect. All businesses will want transactions for their startup, shutdown, and execute methods. I think we can instrument the base `run` method to create transactions for the `startup` and `shutdown` methods, and then leave it to the businesses to start `execute` transactions where they need to.

fajpunk:
> I admit I'm a bit nervous about how Sentry-specific this all is, but I think there's no real way to avoid that and I guess it's just a mass search and replace to switch to any other reporting framework with a similar API.

Yeah, it's definitely Sentry-specific. For the tracing bits, we could use OpenTelemetry instrumentation directly if we wanted to, and maybe that makes more sense for Mobu because it's such a custom integration anyway. But yeah, if we wanted to do that, it should be an easy-medium amount of effort to convert.

> But I'm a bit sad that the context isn't carried by the exceptions themselves, since that would guarantee that anything Python-based would be able to use that data.

We could totally do it with explicitly instrumenting exceptions (via [safir.sentry.SentryException](https://safir.lsst.io/user-guide/sentry.html#special-sentry-exception-types). We could even add `to_sentry` methods to all of the existing exceptions and use all of the existing instrumentation. The global state is confusing in its own way, but to me, it's preferable to [a complicated inheritance graph](https://github.com/lsst-sqre/mobu/pull/374#discussion_r1809078668).  If you and other folks feel differently, I'm glad to implement this a different way.

src/mobu/services/business/base.py
rra: Ah, yeah, that makes sense, it's the same use case as we had before.

src/mobu/services/business/nublado.py
rra: Yup, this looks fine for now.

src/mobu/exceptions.py
rra: So, maybe I care too much about this, but my objection is that we're losing the explanation of what the exception really is in everything *except* Sentry. For example, if we're looking at logs reported to Google (putting aside the problem that we need to fix logging of exceptions in businesses), we would get `KeyError` or something with no context. Likewise in the individual monkey logs.

Even in Sentry, the report to the Slack channel contains the exception type, and `KeyError` or `ValueError` with a possibly useless exception error message feels less helpful than having it be explicit that it's an error creating the TAP client.

I agree that all of the context is in Sentry and you don't need to keep that part, but I was hoping that we could keep the transformation of the exception to something more informative in all those other contexts.

rra:
I think I'm on board with going forward this approach. For mobu in particular, we have a lot of context and I think this is clearly a huge improvement. I'll bring this up during a co-work to talk about where we want to go in the long run, but I think this makes the most sense for at least the next six months or so.

rra (approved):
Everything looks good except that I'll still argue for wrapping the TAP client startup exception for clarity in the systems other than Sentry. But other than that, let's merge this and get a new release out; this is a ton of great work.

src/mobu/exceptions.py
fajpunk: Sold!

PR #399, opened 2025-02-03
Change TAP async to use run_async from pyvo instead of submit_job and polling

stvoutsin:
### Summary

Refactor TAP query execution method to use `pyvo`'s `run_async` instead of manually using `submit_job` and polling to simplify the code and get more accurate query performance.

Downside is that this needs to run synchronously, since we use a run_async is synchronous. If we want to maintain the asynchronous nature we'd have to keep the previous version.

### Tests

Other than running the tests locally, I haven't deployed & tested this in any environment yet.

### DIFF
--- a/src/mobu/services/business/tap.py
+++ b/src/mobu/services/business/tap.py
@@ -99,10 +99,7 @@ class TAPBusiness(Business, Generic[T], metaclass=ABCMeta):
 
                 success = False
                 try:
-                    if self.options.sync:
-                        await self.run_sync_query(query)
-                    else:
-                        await self.run_async_query(query)
+                    await self.run_query(query)
                     success = True
                 finally:
                     await self.events.tap_query.publish(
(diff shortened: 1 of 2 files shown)


src/mobu/services/business/tap.py
```diff
@@ -129,28 +126,17 @@ async def run_async_query(self, query: str) -> None:
...
-        """
-        if not self._client:
-            raise RuntimeError("TAPBusiness startup never ran")
-        self.logger.info(f"Running (sync): {query}")
+        mode = "(sync)" if self.options.sync else "(async)"
+        method = (
+            self._client.search
+            if self.options.sync
+            else self._client.run_async
+        )
```
rra, line 135:
```suggestion
        if self.options.sync:
            mode = "(sync)"
            method = self._client.search
        else:
            mode = "(async)"
            method = self._client.run_async
```

rra (approved):
This is fine -- the old async method wasn't really async anyway since it was making blocking HTTP calls, so it would cause problems for mobu parallelism. At some point we need a library that can make proper Python async calls to TAP as well as use the async UWS methods, but that can wait for another day.

PR #400, opened 2025-02-05
DM-48792: Create new QuerySet runner for SIAv2

stvoutsin:
### Summary 

Enables a new QuerySet runner in mobu for SIAv2 Queries. I've defined as range for ra dec & radius and a fixed time which I know should return results when querying the SIAv2 service. We may want to adjust this, but I don't know what options we want to add and this would require some consultation from Tim or a someone more familiar with the datasets and the available images for given query criteria.
This has been defined in a directory `dp02` under `data/siaquerysetrunner`. Reason for not going with dp0.2 is because I re-use this option to generate the URL for the given dataset, because the design of the SIA API is such that it includes the collection, i.e:

 `api/sia/dp02/query`

### Tests

Only some pytests added here, but have not tested on any environment yet.

### DIFF
--- a/src/mobu/services/business/siaquerysetrunner.py
+++ b/src/mobu/services/business/siaquerysetrunner.py
@@ -1,0 +1,201 @@
+"""Run a set of predefined queries against a SIA service."""
+
+from __future__ import annotations
+
+import asyncio
+import importlib.resources
+from concurrent.futures import ThreadPoolExecutor
+from random import SystemRandom
+from typing import override
+
+import pyvo
+import requests
+import yaml
+from safir.sentry import duration
+from sentry_sdk import set_context
+from structlog.stdlib import BoundLogger
+
+from ...dependencies.config import config_dependency
+from ...events import Events
+from ...events import SIAQuery as SIAQueryEvent
+from ...exceptions import SIAClientError
+from ...models.business.siaquerysetrunner import (
+    SIABusinessData,
+    SIAQuery,
+    SIAQuerySetRunnerOptions,
+)
+from ...models.user import AuthenticatedUser
+from ...sentry import capturing_start_span, start_transaction
+from .base import Business
+
+__all__ = ["SIAQuerySetRunner"]
+
+
+class SIAQuerySetRunner(Business):
+    """Run queries from a predefined set against SIA with random parameters.
+
+    Parameters
+    ----------
+    options
+        Configuration options for the business.
+    user
+        User with their authentication token to use to run the business.
+    events
+        Event publishers.
+    logger
+        Logger to use to report the results of business.
+    flock
+        Flock that is running this business, if it is running in a flock.
+    """
+
+    def __init__(
+        self,
+        *,
+        options: SIAQuerySetRunnerOptions,
+        user: AuthenticatedUser,
+        events: Events,
+        logger: BoundLogger,
+        flock: str | None,
+    ) -> None:
+        super().__init__(
+            options=options,
+            user=user,
+            events=events,
+            logger=logger,
+            flock=flock,
+        )
+        self._running_query: SIAQuery | None = None
+        self._client: pyvo.dal.SIA2Service | None = None
+        self._pool = ThreadPoolExecutor(max_workers=1)
+        self._random = SystemRandom()
+        self.query_set: str = self.options.query_set
+
+        # Load parameters. We don't need jinja here since we're not
+        # generating queries, just parameters from the ranges.
+        params_path = importlib.resources.files("mobu").joinpath(
+            "data", "siaquerysetrunner", self.options.query_set, "params.yaml"
+        )
+        with params_path.open("r") as f:
+            self._params = yaml.safe_load(f)
+
+    @override
+    async def startup(self) -> None:
+        self._client = self._make_client(self.user.token)
+        self.logger.info("Starting SIA Query Set Runner")
+
+    def get_next_query(self) -> SIAQuery:
+        """Get the next SIA query to run.
+
+        Returns
+        -------
+        SIA2SearchParameters
+            SIA query as an SIASearchParameters object.
+        """
+        return self._generate_sia_params()
+
+    @override
+    async def execute(self) -> None:
+        with start_transaction(
+            name=f"{self.name} - execute",
+            op="mobu.sia.search",
+        ):
+            query = self.get_next_query()
+            with capturing_start_span(op="mobu.sia.search") as span:
+                set_context(
+                    "query_info",
+                    {"query": str(query), "started_at": span.start_timestamp},
+                )
+                self._running_query = query
+
+                success = False
+                try:
+                    client = self._client
+                    if not client:
+                        raise RuntimeError(
+                            "SIAQuerySetRunner startup never ran"
+                        )
+                    self.logger.info(f"Running SIA query: {query}")
+                    loop = asyncio.get_event_loop()
+
+                    await loop.run_in_executor(
+                        self._pool,
+                        lambda: client.search(**query.to_pyvo_sia_params()),
+                    )
+                    success = True
+                finally:
+                    await self.events.sia_query.publish(
+                        payload=SIAQueryEvent(
+                            success=success,
+                            duration=duration(span),
+                            **self.common_event_attrs(),
+                        )
+                    )
+
+                self._running_query = None
+                elapsed = duration(span).total_seconds()
+
+            self.logger.info(f"Query finished after {elapsed} seconds")
+
+    @override
+    def dump(self) -> SIABusinessData:
+        return SIABusinessData(
+            running_query=self._running_query, **super().dump().model_dump()
+        )
+
+    def _make_client(self, token: str) -> pyvo.dal.SIA2Service:
+        """Create a SIA client.
+
+        Parameters
+        ----------
+        token
+            User authentication token.
+
+        Returns
+        -------
+        pyvo.dal.SIA2Service
+            SIA2Service client object.
+        """
+        with capturing_start_span(op="make_client"):
+            config = config_dependency.config
+            if not config.environment_url:
+                raise RuntimeError("environment_url not set")
+            sia_url = (
+                f"{str(config.environment_url).rstrip('/')}/api/sia/"
+                f"{self.query_set}"
+            )
+
+            try:
+                s = requests.Session()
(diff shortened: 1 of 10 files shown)


stvoutsin:
Related phalanx branch here:
https://github.com/lsst-sqre/phalanx/tree/tickets/DM-48792

tests/business/siaquerysetrunner_test.py
```diff
@@ -0,0 +1,228 @@
...
+            }
+        ]
+    )
+
+
+@pytest.mark.asyncio
+async def test_random_object(events: Events) -> None:
+    query_set = "dp02"
+    user = AuthenticatedUser(
+        username="bot-mobu-user", scopes=["read:tap"], token="blah blah"
```
fajpunk, line 207: Is `read:tap` the correct scope that a bot user will need to access the SIA service? It doesn't matter in the test, but we'll have to be sure that the correct scope is configured in [the mobu values in phalanx](https://github.com/lsst-sqre/phalanx/blob/tickets%2F48587%2Fmobu-scale-test/applications/mobu/values-idfint.yaml#L48-L52)

src/mobu/models/business/sia.py
```diff
@@ -0,0 +1,54 @@
...
+
+from astropy.time import Time
+from pydantic import BaseModel, Field
+
+from .base import BusinessData, BusinessOptions
+
+__all__ = ["SIA2SearchParameters", "SIABusinessData", "SIABusinessOptions"]
+
+
+class SIA2SearchParameters(BaseModel):
```
fajpunk, line 13: What do you think about `SIA2Query`? It seems like that's what it's representing when it's used

src/mobu/models/business/sia.py
```diff
@@ -0,0 +1,54 @@
...
+        return {"pos": self.pos, "time": times}
+
+    def __str__(self) -> str:
+        """Return a string representation of the query."""
+        times = ", ".join([str(t) for t in self.time])
+        return f"SIA2Query(pos={self.pos}, time=[{times}])"
+
+
+class SIABusinessOptions(BusinessOptions):
+    """Options for any business that runs SIA queries."""
```
fajpunk, line 44: If there are no options here, I don't think we need this subclass, `SIAQuerySetRunnerOptions` can inherit directly from `BusinessOptions`.

src/mobu/services/business/siaquerysetrunner.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Load templates and parameters. The path has to be specified in two
+        # different ways: as a relative path for Jinja's PackageLoader, and as
+        # a sequence of joinpath operations for importlib.resources.
+        template_path = ("data", "siaquerysetrunner", self.options.query_set)
+        self._env = jinja2.Environment(
+            loader=jinja2.PackageLoader("mobu", "/".join(template_path)),
+            undefined=jinja2.StrictUndefined,
+            autoescape=jinja2.select_autoescape(disabled_extensions=["sql"]),
+        )
```
fajpunk, line 64: We don't do any templating right? So we don't need any of the Jinja2 stuff.

src/mobu/services/business/sia.py
```diff
@@ -0,0 +1,163 @@
...
+                try:
+                    if not self._client:
+                        raise RuntimeError("SIABusiness startup never ran")
+                    self.logger.info(f"Running SIA query: {query}")
+                    loop = asyncio.get_event_loop()
+                    await loop.run_in_executor(
+                        self._pool,
+                        self._client.search,
+                        query.to_pyvo_sia2_params(),
+                    )
```
fajpunk, line 110: Is there a way to execute these async, instead of using a threadpool?  I assume no, just double checking

src/mobu/services/business/siaquerysetrunner.py
```diff
@@ -0,0 +1,107 @@
...
+        max_ra = self._params.get("max_ra", 70.0)
+        min_dec = self._params.get("min_dec", -42.0)
+        max_dec = self._params.get("max_dec", -30.0)
+        min_radius = self._params.get("min_radius", 0.01)
+        radius_range = self._params.get("radius_range", 0.04)
+        start_time = self._params.get("start_time", 60550.31803461111)
+        end_time = self._params.get("end_time", 60550.31838182871)
+
+        ra = min_ra + self._random.random() * (max_ra - min_ra)
+        dec = min_dec + self._random.random() * (max_dec - min_dec)
```
fajpunk, line 99: I know these are copied from the TAP businesses, but is there a reason why we're not using `random.uniform` here? Something like `random.uniform(min_ra, max_ra)` ?

src/mobu/models/business/sia.py
```diff
@@ -0,0 +1,54 @@
...
+        -------
+            dict: The query as a dictionary.
+        """
+        times = [Time(str(t), format="mjd").to_datetime() for t in self.time]
+        return {"pos": self.pos, "time": times}
+
+    def __str__(self) -> str:
+        """Return a string representation of the query."""
+        times = ", ".join([str(t) for t in self.time])
+        return f"SIA2Query(pos={self.pos}, time=[{times}])"
```
fajpunk, line 40: If I saw this, I would assume from the formatting that this is a repr of some class somewhere, but it's not. What do you think of something like `SIAv2 parameters: .....`?

fajpunk (commented):
Nice! I've got a couple of stylistic things that you're welcome to discard if you don't agree, and one structural thing:

Are we ever going to want the SIA equivalent of a `TAPQueryRunner` (not `...SetRunner`) business? If not, then I don't think you need to have the `SIABusiness` base class, and you can combine those two business classes.  I can't think of what a `SIAQueryRunner` business would look like (like, what you would put in the config) given that there are no queries like there are for the tap business. It seems that there is only one "query".  But you definitely have a better understanding than I do!

tests/business/siaquerysetrunner_test.py
stvoutsin: Good catch! It's actually `read:image` so I've corrected that bit accordingly

src/mobu/models/business/sia.py
stvoutsin: I think that's what I had initially but then I gave the event the same name and wanted to avoid some confusion there. I changed it back to `SIA2Query`. In terms of the event that was actually `SIAQuery` which I'm now importing as an alias `SIAQueryEvent`. 

Would it make sense to also name that `SIA2Query`  for consistency? Or something like `SIA2QueryEvent`? Or perhaps just leave it as it is now?

src/mobu/services/business/sia.py
stvoutsin: If we use pyvo's SIA search which isn't `async`, that's probably the best we can do for now unless they change it in the future. We could build an asynchronous SIA querying class ourselves that uses something like `aiohttp` but might be a 3-4 days worth of effort kind of thing so I'm not sure if we want to do it now or as an improvement later.

stvoutsin:
> Nice! I've got a couple of stylistic things that you're welcome to discard if you don't agree, and one structural thing:
> 
> Are we ever going to want the SIA equivalent of a `TAPQueryRunner` (not `...SetRunner`) business? If not, then I don't think you need to have the `SIABusiness` base class, and you can combine those two business classes. I can't think of what a `SIAQueryRunner` business would look like (like, what you would put in the config) given that there are no queries like there are for the tap business. It seems that there is only one "query". But you definitely have a better understanding than I do!

Thanks for the suggestions! 

Regarding using a `SIAQueryRunner`, for now I've just removed the SIABusiness base class, I was also a bit confused about what would make sense to be there since SIA does not have an asynchronous mode. My inclination would be to not create an SIAQueryRunner unless we feel like we need it later. If we do I guess it would be because we (or other DFs) would want to specify a list of SIA queries which in this case is no longer just a string (i.e. a query string) but instead a list of SIA2 query parameters (position, time, image format, id etc..). 

Hopefully the changes I've made address all the suggestions, let me know if I've missed anything.

src/mobu/services/business/sia.py
fajpunk: Agreed, not worth the effort now

src/mobu/models/business/sia.py
fajpunk: Ahh yeah I didn't think about the naming conflict with the event. I don't have a super strong opinion, whatever you think is best! For `SIA` vs `SIA2`, I'd say that should be the same in all the names, if I'm understanding correctly.

fajpunk: ...so if this is all for SIA v2, and we're not planning on having a v1 business, I'm totally fine with leaving the `2` off.

fajpunk:
>  My inclination would be to not create an SIAQueryRunner unless we feel like we need it later.
Makes sense.  In that case, what do you think about absorbing the `services.business.sia.SIABusiness` class into the `services.business.siaquerysetrunner.SIAQuerySetRunner` class, given that there's not likely going to be any other subclasses of `services.business.sia.SIABusiness`?

Similarly, what about moving the models in `models.buisiness.sia` into `models.business.siaquerysetrunner`?

src/mobu/models/business/sia.py
stvoutsin: Ok makes sense. Since I don't expect we'd be running an SIA v1 query I've removed all instances of the version number except for one place, which is the external pyvo `SIA2Service` class.

stvoutsin:
> Makes sense. In that case, what do you think about absorbing the `services.business.sia.SIABusiness` class into the `services.business.siaquerysetrunner.SIAQuerySetRunner` class, given that there's not likely going to be any other subclasses of `services.business.sia.SIABusiness`?
> 
> Similarly, what about moving the models in `models.buisiness.sia` into `models.business.siaquerysetrunner`?

Sure that sounds reasonable to keep things simpler, I've merged in the model & service from sia into siaquerysetrunner.

fajpunk:
I was thinking of getting rid of the base service class entirely, something like: https://github.com/lsst-sqre/mobu/pull/428/commits/b1d5f5b688b4b5368fefaaf92d3a42c35181cdaa

What do you think?

stvoutsin:
> I was thinking of getting rid of the base service class entirely, something like: #428 What do you think?

Hm yea I think you're right, let's keep it simple for now and if we want to add another SIA Business we can add an abstraction layer. I'll update the branch with your recommendation

fajpunk (approved):

PR #401, opened 2025-02-06
DM-48825: Properly close async generators

rra:
Change the type of functions returning async generators to `AsyncGenerator` and properly close async generators in the hope that this will address various log errors about generators already running.


fajpunk (approved):
:crossed_fingers:

PR #403, opened 2025-02-08
DM-48838: Create Gafaelfawr tokens in parallel

rra:
When creating a flock, create the Gafaelfawr tokens in parallel (up to the limit of the httpx connection pool, which is 100) instead of serially.

No business is using the shared global httpx client and connection pool, so remove it from the constructors of all of the business classes and stop passing it down. The Nublado business uses a separate connection pool per monkey maintained in the Nublado client, and the TAP business uses pyvo, which manages its own connections.


fajpunk (approved):
Do we also want to add a parameter to the Safir http client to control the connection pool size?

PR #407, opened 2025-02-14
DM-48941: Only check out notebook repos once

fajpunk:
Add a process-global notebook repo cache. Use this in the NotebookRunner business to only clone notebook repos once per process, not once per monkey.

The notebook repo refresh functionality complicates this a bit, see [this comment](https://github.com/lsst-sqre/mobu/pull/407#issue-2854417705)

### DIFF
--- a/src/mobu/services/business/notebookrunner.py
+++ b/src/mobu/services/business/notebookrunner.py
@@ -9,12 +9,10 @@ from __future__ import annotations
 import contextlib
 import json
 import random
-import shutil
 from collections.abc import AsyncGenerator, Iterator
 from contextlib import asynccontextmanager
 from datetime import timedelta
 from pathlib import Path
-from tempfile import TemporaryDirectory
 from typing import Any, override
 
 import sentry_sdk
@@ -46,7 +44,7 @@ from ...models.repo import RepoConfig
 from ...models.user import AuthenticatedUser
 from ...sentry import capturing_start_span, start_transaction
 from ...services.business.base import CommonEventAttrs
-from ...storage.git import Git
+from ...services.repo import RepoManager
 from .nublado import NubladoBusiness
 
 __all__ = ["NotebookRunner"]
@@ -83,6 +81,7 @@ class NotebookRunner(NubladoBusiness):
         *,
         options: NotebookRunnerOptions | ListNotebookRunnerOptions,
         user: AuthenticatedUser,
+        repo_manager: RepoManager,
         events: Events,
         logger: BoundLogger,
         flock: str | None,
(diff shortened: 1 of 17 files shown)


src/mobu/services/business/notebookrunner.py
```diff
@@ -116,22 +117,40 @@ async def startup(self) -> None:
...
+            mkdtemp(prefix=f"mobu-notebooks-{self.user.username}-")
+        )
+        self._repo_hash = info.hash
 
-        repo_config_path = self._repo_dir / GITHUB_REPO_CONFIG_PATH
+        # Each running business needs its own copy of the notebooks because a
+        # github refresh webhook may come in at any time and delete the repo
+        # manager's copy of the repo. We'll then get an error when we try to
+        # read those files.
+        copytree(src=info.path, dst=self._repo_path, dirs_exist_ok=True)
```
fajpunk, line 151: Other possible ways around this:
* Just don't ever delete old clones after a refresh. This would be an intentional storage leak.
* Coordinate access to the repo cache among all of the monkeys so that the monkeys know to not try to read any more notebooks until a given repo has been refreshed. This seems difficult.

src/mobu/services/repo_cache.py
```diff
@@ -0,0 +1,114 @@
...
+            The git ref to checkout after the repo is cloned
+        hash
+            The hash of the cloned repo to remove
+        """
+        key = self._make_key(url=url, ref=ref)
+
+        # This theoretically doesn't need a lock, but if we add any awaits here
+        # in the future, we'd need to add a lock, and it would be pretty easy
+        # to forget to do it.
+        async with self._remove_lock:
```
rra, line 103: It feels weird to me to have two separate locks here instead of one protecting all manipulations of the underlying file system. I *think* this is safe, or at least I can't construct a scenario where it's a problem, but I think ti's more obvious if you use one lock, and I don't think it will matter much.

rra (approved):
Nothing major here and I think the overall approach is fine as is, but as noted I would be inclined to never delete checkouts for now and implement reference counting in the long term.

src/mobu/services/repo.py
```diff
@@ -0,0 +1,164 @@
...
+        reference = _Reference(url=url, ref=ref, hash=repo_hash)
+
+        # This theoretically doesn't need a lock, but if we add any awaits here
+        # in the future, we'd need to add a lock, and it would be pretty easy
+        # to forget to do it.
+        async with self._lock:
+            info = self._cache.get(key)
+            if info:
+                logger.info("Invalidating repo")
+                del self._cache[key]
```
rra, line 152: Hm, I think this means that if any monkey shuts down, it invalidates the cache for its repository and forces a reclone even if that ends up being the same hash. I think this is fine for now because doing something more complicated requires adding the concept of checking to see if a given checkout of a repo is still current, but it might be worth a comment saying that this will result in some unnecessary clones still.

rra (approved):
One minor note about commenting some behavior, but otherwise this looks great to me. Thanks!

PR #410, opened 2025-02-17
DM-48977: Switch to new Python type variable syntax

rra:
Switch to the new Python syntax for generic parameters to classes and functions. This apparently unconfused mypy and Ruff enough that it uncovered some additional diagnostics, which required some `noqa` comments and some code refactoring, one to fix a bug and another to unconfuse the type checker.

### DIFF
--- a/src/mobu/services/business/base.py
+++ b/src/mobu/services/business/base.py
@@ -8,7 +8,7 @@ from asyncio import Queue, QueueEmpty
 from collections.abc import AsyncGenerator, AsyncIterable
 from datetime import timedelta
 from enum import Enum
-from typing import Generic, TypedDict, TypeVar
+from typing import TypedDict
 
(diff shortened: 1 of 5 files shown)


src/mobu/services/business/base.py
```diff
@@ -114,21 +111,21 @@ def __init__(
 
     # Methods that should be overridden by child classes if needed.
 
-    async def startup(self) -> None:
+    async def startup(self) -> None:  # noqa: B027
```
fajpunk, line 114: What about `return None` for these instead of disabling linting? That is, if that gets rid of the warning which I don't actually know if it does...

rra: Oh, good call, I will try that and see if that suppresses the Ruff warning.

PR #414, opened 2025-02-18
DM-48977: Move dependencies to pyproject.toml

rra:
Make mobu look a bit more like a standard Python project by moving the runtime dependencies to `pyproject.toml`. Continue to generate frozen dependencies for the Docker build as before, and continue to handle development dependencies as before.


fajpunk:
What do you think about moving the dev deps into pyproject.toml as well, with dependency groups:
https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies

rra:
I would absolutely love to use that stuff but when I first looked at it, it didn't work with tox-uv. If it works with tox-uv now, and not just with the single venv in the `.venv` directory, then we absolutely should adopt it. I kept waiting for that to show up in the change log of a major release, but it's quite possible that they fixed it in a non-major release since it wouldn't have broken backward compatibility and I just missed it.

PR #426, opened 2025-02-18
DM-49009: Stagered monkey start

fajpunk:
Add two new flock config parameters to let us start monkeys in batches with waits in between:
* `start_batch_size`: the number of monkeys to start in each batch
* `start_batch_wait`: the amount of time to wait in between starting batches


rra (approved):
Oh, that's a nice way of implementing this. I like that.

PR #433, opened 2025-02-24
DM-49117: CI - Run all notebooks in PR

fajpunk:
Not just the ones changed in the latest commit. Fixes this situation:

* PR with notebooks `a`, and `b` gets submitted
* mobu check fails with errors in notebooks `a` and `b`
* A new commit that fixes `a` but NOT `b` is pushed
* mobu check runs ONLY with `a` because it's the only notebook that was changed in the most recent commit
* mobu check passes because `a` is fixed... but `b` is still broken
* GitHub will say everything is good with the PR

Needs https://github.com/lsst-sqre/safir/pull/387


rra (approved):
Nice, good fix.

PR #436, opened 2025-02-27
DM-49241: Record full repo-relative notebook path on events

fajpunk:
We're currently recording just the notebook filename in the `notebook` tag on events. We should be recording the entire path relative to the root of the repo.


rra (approved):
Looks good to me!

PR #438, opened 2025-03-05
DM-49319: Configurable gafaelfawr http client timeout

fajpunk:
Add a `gafaelfawr_timeout` config option. With very large numbers of users, like for scale testing, the default httpx timeouts from the [safir http client](https://safir.lsst.io/user-guide/http-client.html) may not be long enough.


rra (approved):
Looks good to me! Once this is merged, I'll also give you a PR to batch the user creation a bit more sensibly.

PR #442, opened 2025-03-07
DM-49374: Batch Gafaelfawr token creations

rra:
Batch the creation of Gafaelfawr tokens during flock startup and only perform 10 at a time. This will still provide some parallel speed-up without causing a pile-up on the Gafaelfawr side waiting for database transactions, which in turn causes problems with HTTP timeouts on the mobu side.


fajpunk (approved):
Do you think we still need the gafaelfawr timeout config option?

rra:
Maybe not, but it can't hurt. I think we can keep it; it's a fairly small change and it might come in handy.

PR #444, opened 2025-03-10
DM-49388: Add support for Nublado with user subdomains

rra:
Update to the new version of the Nublado client with user subdomain support and configure the test suite to run all Nublado tests both with and without subdomains to ensure that mobu doesn't make any assumptions that break with user subdomains.


tests/business/emptyloop_test.py
```diff
@@ -58,7 +58,7 @@ async def test_run(
     }
 
     # Check events
-    published = cast(MockEventPublisher, events.empty_loop).published
+    published = cast("MockEventPublisher", events.empty_loop).published
```
fajpunk, line 61: Why change this? I assume that the typechecker still knows the interface if we do, though I don't know how. And if we do, does it mean we should get rid of the `MockEventPublisher` import?

tests/conftest.py
```diff
@@ -216,15 +192,19 @@ async def anon_client(app: FastAPI) -> AsyncGenerator[AsyncClient]:
         yield client
 
 
-@pytest.fixture
+@pytest.fixture(ids=["shared", "subdomain"], params=[False, True])
```
fajpunk, line 195: Oh neat, I didn't know fixtures could be parameterized!

tests/business/emptyloop_test.py
rra: This is just a Ruff thing. The latest version of Ruff always wants the first argument to `cast` to be quoted. See [TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/). I'm not sure I completely buy the logic, but it doesn't seem worth overriding since it doesn't really matter.

tests/conftest.py
rra: Yeah, I didn't either until I was trying to figure out how to do this, and then it took me a while to figure out how to assign nice names to the options, but it's a very cool feature and very magical!

fajpunk (approved):

PR #448, opened 2025-03-18
DM-49361: Multi-instance mode

fajpunk:
## New features

- Multiple replicas of mobu can be run in an environment. Note that this comes with certain restrictions, see the [docs changes in this PR](https://github.com/lsst-sqre/mobu/blob/0ff58060c23f8c5ee5cce0f5dd08a4312ba50562/docs/user-guide/multiple-replicas.rst) for more details.

## Bug fixes

- One period `execution_idle_time` elapses after trying to execute a notebook with no code cells, instead of not waiting at all.


src/mobu/services/business/notebookrunner.py
```diff
@@ -272,14 +272,22 @@ async def execute_notebook(
...
         ) as span:
             try:
-                for cell in self.read_notebook(self._notebook):
+                cells = self.read_notebook(self._notebook)
+
+                # We want to wait if the notebook is totally empty so we don't
+                # spin out of control on empty notebooks
+                logger.warning("empty notebook")
+                if not cells:
+                    await self.execution_idle()
```
fajpunk, line 289: This will probably get us to our load test goal without multiple replicas :smile: :cry:

src/mobu/services/flock.py
```diff
@@ -209,11 +217,22 @@ def _create_monkey(self, user: AuthenticatedUser) -> Monkey:
     async def _create_users(self) -> list[AuthenticatedUser]:
         """Create the authenticated users the monkeys will run as."""
         users = self._config.users
+        if users and self._replica_count > 1:
+            raise RuntimeError(
+                "Only user_spec users are allowed if replica_count > 1"
+            )
         if not users:
             if not self._config.user_spec:
                 raise RuntimeError("Neither users nor user_spec set")
```
fajpunk, line 226: Actually, any objections to moving all of this validation into the config model? We know it all at config time.

docs/user-guide/multiple-replicas.rst
```diff
@@ -0,0 +1,49 @@
...
+Mobu should currently only be run with multiple replicas during explicitly monitored periods, like scheduled load testing.
+Otherwise, the behavior described above could lead to confusion.
+If other use cases for multiple replicas come up in the future, we can do some more work to mitigate some of these downsides.
+
+Differences From a Single Replica
+---------------------------------
+
+The only difference when running as one of multiple replicas is that ``user_spec`` usernames will be further prefixed with an instance identifier.
+Mobu adds this prefix if the ``MOBU_REPLICA_COUNT`` env var contains a value that is greater than 1.
+This instance identifier comes from the ``MOBU_INSTANCE_ID`` env var.
```
rra, line 29: The thing that makes me a little unhappy about this is that we generate yet another different set of usernames and thus burn yet another full block of UIDs when running in this mode.

I think the multiple instance stuff only really works with autostart configuration. (If one talks to the API, then one is going to start that flock on whatever instance happened to service the autostart request and there's no way to divide it across replicas.) So how about this alternate proposal: If the replica count is greater than 1, every instance knows its own sequence number in a range from 0 to one less than the replica count. Each instance then generates the list of monkeys that it would run and, for each sequence in that list, mods it by the replica count. If the result of the mod operation equals its replica sequence number, then it runs that monkey; otherwise, it doesn't.

This basically shards the monkeys across each of the mobu replicas but keeps the same username patterns so that we don't generate new UIDs, home directories, etc., when we shard.

src/mobu/services/flock.py
```diff
@@ -209,11 +217,22 @@ def _create_monkey(self, user: AuthenticatedUser) -> Monkey:
     async def _create_users(self) -> list[AuthenticatedUser]:
         """Create the authenticated users the monkeys will run as."""
         users = self._config.users
+        if users and self._replica_count > 1:
+            raise RuntimeError(
+                "Only user_spec users are allowed if replica_count > 1"
+            )
```
rra, line 223: This should probably raise a custom exception that inherits from Safir's `ClientRequestError` so that mobu returns a 400 error if someone tries to create a flock rather than a 500 error. Or, alternately, only pass in the replica count and instance ID for autostart configurations and don't use them at all when the user manually creates a flock via the API, since they don't really make sense in that case anyway unless we're going to add Redis or something for runtime coordination of spreading monkeys across flocks (which feels like total overkill).

docs/user-guide/multiple-replicas.rst
fajpunk: > I think the multiple instance stuff only really works with autostart configuration.
Agreed.

A couple of other options:
1. Assuming you're suggesting we keep the way for an instance knowing its sequence number (injecting a `StatefulSet` index), then we could change `MOBU_INSTANCE_ID` to `MOBU_SEQUENCE_NUMBER`, and then just multiple the the generated user index by the sequence number (+ 1), right?
1. Keep the logic here and ALWAYS add the instance id, and burn a set of uids just once in every env when we roll this out.

Out of those two options, I think I like the first (a slightly simpler version of yours).

src/mobu/services/flock.py
fajpunk: I like raising a `ClientRequestError`. 

I considered disabling this handler, and other functionality that may be wonky, like the GitHub functionality. I decided it's not worth it for now and added a bunch of warnings to the docs and Phalanx config instead, but I could definitely be convinced otherwise.

docs/user-guide/multiple-replicas.rst
rra: I'm gathering that you don't like the idea of sharding the users instead? One of the things that confused me about this implementation is that I think it means each mobu would independently start N monkeys, right? So basically the monkey count of everything is multiplied by the replica count? That was the part that felt semantically weird to me; I would expect each mobu instance to instead run only its shard of the users for all of the configured flocks, so the size of the flock is accurate and is just divided across the mobu replicas.

Maybe I have a fundamental misunderstanding here, though.

src/mobu/services/flock.py
rra: That's fine with me; we can always clean it up later.

docs/user-guide/multiple-replicas.rst
fajpunk: TLDR: I'm glad to do the sharding, I'll change it.

> One of the things that confused me about this implementation is that I think it means each mobu would independently start N monkeys, right?

Yep, [exactly](https://github.com/lsst-sqre/mobu/pull/448/files#diff-c6ec1762e456b9bae3fde30289b02767fc27ca8b79f0df986b1bc8f4cd93d5b5R33-R35)!

I get it now, I misunderstood. I started with the sharding like you're proposing, where `count` was the total number of monkeys across all of the instances, but changed it so that the entire config would be applicable to a single instance, because it seemed weird to me that that one particular thing would change the number of monkeys in each instance depending on the replica count, but now I can see how it seems weird the other way. Glad to do the sharding.

fajpunk: As an added bonus, this also gets rid of any restrictions on how to specify users.

src/mobu/services/flock.py
fajpunk: Sharding the monkeys across replicas gets rid of any of these restrictions on specifying users, so this doesn't apply at all any more.

rra (approved):
This looks great. Thank you!

PR #453, opened 2025-06-06
DM-51291: Simpler, more flexible notebook filtering

fajpunk:
Make notebook filtering simpler and more flexible.

* Deprecate `exclude_dirs`
* Add `collection_rules` to every notebookrunner business, not just `NotebookRunnerList`

`collection_rules` looks like this:

```yaml
   collection_rules:
     - type: "exclude_union_of"
       patterns:
         - "not-these/**"
         - "not/these/either/**"
     - type: "intersect_union_of"
       patterns:
         - "this.ipynb"
         - "these/**"
         - "also/these**"
     - type: "include"
       patterns:
         - "**/these-*"
```

Each entry is a pattern using the `Python pathlib glob pattern language`_.

* Start with all notebooks in the repo.
* For each collection rule, remove notebooks:
  * Intersect rules will remove notebooks that are not in the
    intersection of:
      * The current set
      * The union of the matched patterns.
  * Exclude rules will remove notebooks from the current set that are
    in the union of the matched patterns.
* Remove any remaining notebooks that require unavailable services.

The include rules logic may sound complex, and I agree. I'm definitely open to suggestions on better ways to do this.  It is there to support the ability to specify rules that say that notebooks should match ANY pattern (a single include rule with multiple patterns), and rules that say that notebooks should match ALL patterns (multiple include rules).  Multiple include rules will likely be used rarely in flock specs or in-repo config, but the functionality is used to further filter notebooks in CI jobs, beyond what any in-repo config may have already filtered.

### DIFF
--- a/changelog.d/20250617_170359_danfuchs_HEAD.md
+++ b/changelog.d/20250617_170359_danfuchs_HEAD.md
@@ -9,25 +9,28 @@
 
   ```yaml
      collection_rules:
-       - type: "exclude"
+       - type: "exclude_union_of"
          patterns:
            - "not-these/**"
            - "not/these/either/**"
-       - type: "include"
+       - type: "intersect_union_of"
          patterns:
            - "this.ipynb"
            - "these/**"
            - "also/these**"
-       - type: "include"
+       - type: "intersect_union_of"
          patterns:
            - "**/these-*"
   ```
 
   Each entry is a pattern using the [Python pathlib glob pattern language](https://docs.python.org/3/library/pathlib.html#pathlib-pattern-language)
 
-  * If no include rules are specified, start with all notebooks.
-  * For every include rule:
-    * Get the set of all of the notebooks matched by any pattern.
-  * Take the intersection of all of those sets.
-  * Subtract any notebook matched by an exclude pattern.
-  * Subtract any notebook that specifies required services that are missing.
+* Start with all notebooks in the repo.
+* For each collection rule, remove notebooks:
+  * Intersect rules will remove notebooks that are not in the
+    intersection of:
+      * The current set
+      * The union of the matched patterns.
+  * Exclude rules will remove notebooks from the current set that are
+    in the union of the matched patterns.
+* Remove any remaining notebooks that require unavailable services.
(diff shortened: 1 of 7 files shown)


rra (approved):
This looks good overall. The only concern that I have is the syntax of how it's expressed in the configuration file, because I never would have guessed the meaning of the configuration if I were just looking at the configuration file without the documentation. My assumption, since it's a list of rules, is that those rules would be applied in order:

- Start with all or no notebooks depending on whether the first rule is an include or exclude
- As each include rule is encountered, add those notebooks
- As each exclude rule is encountered, remove those notebooks

This is not at all what the config does and indeed it looks like the order of the rules (and particularly the order of the includes relative to the excludes) is meaningless. If I'm right about the intention, I'd love to have the rules be expressed in some way that makes this more obvious. Maybe something like:

```yaml
include:
    intersection:
        - union:
            - path/**.ipynb
        - union:
            - **/test_*
```
That's probably not it exactly, but maybe it provides the idea? Have separate include and exclude blocks (since their order doesn't matter) rather than a list of rules, but inside the include block express the union and intersection explicitly somehow.

fajpunk:
I just changed the names of the rules to be more clear, and changed it to work so that it does in fact work like you (both you specifically and the royal "you") think it does from the config:

```
 collection_rules:
     - type: "intersect_union_of"
       patterns:
         - "not-these/**"
         - "not/these/either/**"
     - type: "intersect_union_of"
       patterns:
         - "this.ipynb"
         - "these/**"
         - "also/these**"
     - type: "include"
       patterns:
         - "**/these-*"
```
Another option is just to have `include_patterns` and `exclude_patterns`, have a special case CI runner that overrides `find_notebooks()` and not support multiple intersections :shrug:

What do you think?

rra:
I'm confused, but maybe I'm just missing something. The current code doesn't seem to match your comment; it just has `exclude` and `include` blocks in a list. And I think it still has the property that confused me, namely that there's a list of intermixed `include` and `exclude` blocks, but the order doesn't matter, so there doesn't seem to be any reason for it to be a list, as opposed to just two keys, one `include` and one `exclude`, that take a list of patterns each.

Did I just misunderstand this completely somehow?

fajpunk:
Gah sorry @rra, I didn't not push the commit that did all of the stuff I said I did. It is now pushed.

PR #461, opened 2025-07-01
DM-51647: Support adding users to groups

rra:
Some RSP services, such as Butler, now use group membership for access control and therefore can only be tested by users in appropriate groups. Support setting group membership for users and sending that to Gafaelfawr when creating the mobu monkey token.


fajpunk (approved):
Thank you for slogging through all the tests :vomiting_face: !

rra:
Yeah, I really need to figure out how to refactor the tests so that changes to mobu are less irritating. I'm convinced there's a better abstraction hiding in there somewhere.

PR #465, opened 2025-07-09
DM-51751: Run CI jobs on pull_request events instead of `check_suite/run`

fajpunk:
There is a bug where a Mobu CI job will never run on a PR if there is
only one commit on the branch in the PR:

* User pushes a branch with some commits in it
* Mobu gets a `check_suite` event, sees that the associated commit is
  not associated with a PR, doesn't start a job
* User opens a PR, which doesn't send any trigger any GitHub Mobu CI app
  webhook events.

At least one notebook repo
([tutorial-notebooks](https://github.com/lsst/tutorial-notebooks)) is
configured to block merges unless the Mobu CI job passes. If a user does
not add any more commits to the PR branch, then the mobu job will never
be run, and the PR will be forever blocked (or blocked until the user pushes an
empty commit, which is the current workaround).

Since we don't ever want to run Mobu CI on commits that are not in a PR,
Mobu can listen for PR events instead. This makes the Mobu webhook
handlers slightly simpler as an added bonus.


jonathansick (approved):
Looks good. I agree just checking on PR events is the right way to go here.

PR #530, opened 2026-06-01
DM-55115: Update the Nublado client version and add timeouts

rra:
Update the version of the Nublado client to speak the new Nublado WebSocket protocol, which fixes messaging issues with a JupyterLab with jupyter-server-documents installed.

Add a new `timeout` configuration option for the `NubladoPythonLoop` business that enforces an execution timeout each time the code is run.

Add a new `cell_execution_timeout` configuration option for the notebook runner businesses that enforces a per-cell execution timeout. The same timeout is used for all cells, so it must be longer than the expected execution time of any cell. Its primary intended use is as a backstop to ensure that a cell cannot hang forever and thus block the monkey without producing an error. Setting it to an hour or two is often reasonable.

Enforce a timeout of one minute per Python block on the probe Python code run after starting a new Nublado notebook to gather information about that notebook and set up the working directory. This catches problems where the JupyterLab kernel ignores the code execution message, which would otherwise cause the monkey to silently hang.

### DIFF
--- a/src/mobu/services/business/notebookrunner.py
+++ b/src/mobu/services/business/notebookrunner.py
@@ -15,14 +15,8 @@ from datetime import timedelta
 from pathlib import Path
 from typing import Any, override
 
-import sentry_sdk
 import yaml
-from rubin.nublado.client import (
-    CodeContext,
-    JupyterLabSession,
-    NubladoExecutionError,
-    NubladoExecutionTimeoutError,
-)
+from rubin.nublado.client import CodeContext, JupyterLabSession
 from rubin.repertoire import DiscoveryClient
 from safir.sentry import duration
 from sentry_sdk import set_context, set_tag
@@ -32,11 +26,7 @@ from structlog.stdlib import BoundLogger
 from ...constants import GITHUB_REPO_CONFIG_PATH
 from ...dependencies.config import config_dependency
 from ...events import Events, NotebookCellExecution, NotebookExecution
-from ...exceptions import (
-    NotebookCellExecutionError,
-    NotebookRepositoryError,
-    RepositoryConfigError,
-)
+from ...exceptions import NotebookRepositoryError, RepositoryConfigError
 from ...models.business.notebookrunner import (
     NotebookRunnerData,
     NotebookRunnerOptions,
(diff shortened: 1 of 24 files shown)


src/mobu/services/business/notebookrunner.py
```diff
@@ -400,7 +404,10 @@ async def execute_cell(
                 )
 
                 notebook = getattr(context, "notebook", "<unknown notebook")
-                msg = f"{notebook}: Error executing cell"
+                if isinstance(e, NubladoExecutionTimeoutError):
+                    msg = f"{notebook}: Timeout executing cell"
+                else:
+                    msg = f"{notebook}: Error executing cell"
```
fajpunk: I'm 100% in favor of putting as much as we can into structured attributes for logging, so maybe move the notebook into an attribute here... assuming it isn't already, and in that case, you can take it out of the message.

tests/business/nubladopythonloop_test.py
```diff
@@ -892,3 +893,36 @@ async def test_user_spec_groups(
...
+async def test_code_timeout(
+    *,
+    data: Data,
+    client: AsyncClient,
+    mock_jupyter: MockJupyter,
+    sentry_items: Captured,
+) -> None:
+    config = data.read_json("solitary/input/python-timeout")
+    code = config["business"]["options"]["code"]
+    mock_jupyter.register_python_result(code, "4", delay=timedelta(seconds=2))
```
fajpunk: This is ironic coming from me, as I've probably added 30s to the test suite execution time, but is it worth making these of 1s (or even less) and declaring timeouts of .5s (or even less)? As short as we can get away with such that the computer can still consistently make the mock execution longer than the specified timeout?

fajpunk (approved):
Finally! This is great, and I'm totally in favor of splitting all the log messages we can into structured attributes.

tests/business/nubladopythonloop_test.py
rra: Ah, right, `HumanTimedelta` can take a float, so we can use a subsecond timeout. Thank you!

src/mobu/services/business/notebookrunner.py
rra: I'm not sure why we have a separate `NotebookCellExecutionError` exception at all, since the Nublado client already returns a rich exception with Sentry annotations. Unless you remember a reason why we needed another exception layer, I'm inclined to delete this whole block and just propagate the underlying exception, since it should already have all of the relevant annotations from `context`.

fajpunk: I'm good getting rid of this layer of exceptions. I think this was from me over-extrapolating [this comment](https://github.com/lsst-sqre/mobu/pull/391#discussion_r1926229791), to be very explicit about what phase of execution the error happened in the case where the error isn't the Nublado client code execution error.

fajpunk: Seems that direct link to the comment doesn't work:

> That try/catch block is there from hard-won experience, I fear. This change will cause it to throw random things like ValueError and KeyError with absolutely no useful information about what precisely failed. It's there to get the context of "this happened while creating a TAP client" into the exception along with the underlying exception type and message.

rra: Ah, I think that doesn't apply to the Nublado exceptions specifically but it probably does still apply to other possible failures. Let me see if I can find a good way of handling that.

PR #531, opened 2026-06-01
Pin back jupyter_server to < 2.18

athornton:


rra:
Should be resolved by the just-merged PR.

